PR #2, opened 2017-08-25
Added cohort collapse functionality.

chrisknoll:
Contribution from @gowthamrao

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>org.ohdsi</groupId>
     <artifactId>circe</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>1.1.0-SNAPSHOT</version>
 
     <properties>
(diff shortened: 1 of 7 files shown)


chrisknoll:
@gowthamrao: can you do a quick review on these changes: I copped them over from the WebAPI merge we just had, but those files are going to be removed when we start referencing the stand-alone library for cohort building.

2 Changes you should note:
- I swtiched CollapseType to be an enum with one value 'ERA'.  This will make it simpler to check for a collapse type.
- I must have missed this in the final PR, but the collapseSettings was hard-coded to Integer(0) instead of reading it from the CollapseSettings.  So that change was made here, but not back in WebAPI.  We won't need to change it in WebAPI after this libary becomes referenced in webapi.

Let me know if you have any concerns. I'd like to merge this in today, sorry for the short notice.

-Chris

PR #7, opened 2017-09-12
Events within visit

gowthamrao:
This pull-request addresses the following:

1. New major functionality in creating Qualified_events. Option to restrict events within same visit
2. visit_occurrence is maybe used to generate statistics about cohort-creation
3. changed row_number() to dense_rank() to allow to restrict to earliest-event, latest-event or all-event in qualified-event queries.
https://github.com/OHDSI/Atlas/issues/454

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
@@ -541,6 +541,12 @@ public class CohortExpressionQueryBuilder implements IGetCriteriaSqlDispatcher, 
 
       clauses.add(String.format("A.END_DATE >= %s AND A.END_DATE <= %s", startExpression, endExpression));    
     }
+	
+	// RestrictVisit
+	boolean restrictVisit = corelatedCriteria.restrictVisit;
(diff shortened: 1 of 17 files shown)


gowthamrao:
@chrisknoll - please let us know if you find any issues with this PR

chrisknoll:
Will do. I'm sorry I haven't been able to turn this around sooner, there's been other priorities I've had to manage. I'll let you know if there's any issues by 9/20.

chrisknoll:
@gowthamrao , as an aside, and I don't want to hold this PR up, but the change you made regarding dense_rank() wasn't made in all places.  You did it when the primary events were selected, but you didn't put it in to the individual criteria queries such as here:
https://github.com/OHDSI/circe-be/blob/master/src/main/resources/resources/cohortdefinition/sql/conditionOccurrence.sql#L5

Pretty much anywhere you see 'as ordinal' in the select, you should use dense_rank().  The reason for this is that when you make a correlated query, you want to be able to find events that occur on the same day when they choose 'first event'.  But this is a minor change and shouldn't hold up this PR here, so I'm merging in this to get the 'same visit' criteria implemented.

Next step will be to modify the Atlas circe ui editor to provide the checkbox to let people restrict to same visit.

PR #8, opened 2017-09-21
row_number() to dense_rank()

gowthamrao:
Changed row_number() to dense_rank() everywhere ordinal was created for use in restriction to earliest_event or latest_event.

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CorelatedCriteria.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CorelatedCriteria.java
@@ -37,6 +37,6 @@ public class CorelatedCriteria {
   @JsonProperty("Occurrence")
   public Occurrence occurrence;
   
-  @JsonProperty("restrictVisit")
+  @JsonProperty("RestrictVisit")
   public boolean restrictVisit = false;
 }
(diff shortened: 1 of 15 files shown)


src/main/java/org/ohdsi/circe/cohortdefinition/CorelatedCriteria.java
```diff
@@ -37,6 +37,6 @@
   @JsonProperty("Occurrence")
   public Occurrence occurrence;
   
-  @JsonProperty("restrictVisit")
+  @JsonProperty("RestrictVisit")
```
chrisknoll: Thank you for changing this to follow the other case conventions of the criteria objects.  While this doesn't follow java naming conventions, it is what was established, and we would have difficulty changing all the other properties to be camelCase due to existing data that is out three. So this is a good fix to find now before it becomes widely available.

chrisknoll:
@gowthamrao : This will merge into master, but using squash to combine those commits into a single one.  This means that your upstream fork of the master branch may have problems syncing.  In the future, you should create a feature branch off master to submit off your PR. Once the PR is accepted, you'll see the change on the upstream master branch which you sync down to your local fork.

PR #10, opened 2017-10-03
Revert dense_rank() to row_number() to enforce one event per person.

### LINKED ISSUE #9: Reporting confusion between row_number() and dense_rank()
@gowthamrao ,

This is related to some of the recent work that changed the selecting 'earliest' and 'first' as using dense_rank() as the window function, allowing multiple events on the same day to be evaluated.

The problem stems from the 'inclusion report' statistics where prior to dense_rank(), the report was able to be viewed as a 'person-level' report because row_number() means that a single event would be returned per person.  Now that we use dense_rank(), we run into the case where if a person has 5 events of a type in the same day, the inclusion report appears like there were 5 people evaluated.   In the row_number() case, it would only use 1 event, and then the count in the report would be 1, which would correspond to a count of people.  @pbr6cornell  has concerns (as do I) about how this change means a breaking change with how the report numbers are interpreted.

This issue report is to discuss how we can think about how we specify the cohort definition in a way that we can talk in terms of 1 event per person so that our reporting can be more interpret-able.  Some ideas:

- use row_number() in the 'earliest per person/latest per person' but leave dense_rank() in the 'first occurrence in history,  
- Have a global cohort definition setting that will specify if we should use row_number() or dense_rank() (not sure how we would word this)
- Change up our reporting statistics so that it becomes something more person-centric

This isn't all the options, but things to think about.  I feel that we are going to need to come up with a solution for this quickly because the assumption that going to dense_rank() is going to have very large negative effects on the people that interpret the reports, and so we're going to want to address this prior to it's inclusion into next release of atlas.

chrisknoll:
Updated library to v1.2.1.

Fixes #9.

### DIFF
--- a/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
+++ b/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
@@ -8,7 +8,7 @@ SELECT event_id, person_id, start_date, end_date, op_start_date, op_end_date, vi
 INTO #qualified_events
 FROM 
 (
-  select pe.event_id, pe.person_id, pe.start_date, pe.end_date, pe.op_start_date, pe.op_end_date, dense_rank() over (partition by pe.person_id order by pe.start_date @QualifiedEventSort) as ordinal, pe.visit_occurrence_id
+  select pe.event_id, pe.person_id, pe.start_date, pe.end_date, pe.op_start_date, pe.op_end_date, row_number() over (partition by pe.person_id order by pe.start_date @QualifiedEventSort) as ordinal, pe.visit_occurrence_id
   FROM primary_events pe
   @additionalCriteriaQuery
 ) QE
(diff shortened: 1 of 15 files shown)


src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
```diff
@@ -56,13 +56,11 @@ from #included_events;
...
 @censoringInserts
 
-
-
 with collapse_constructor_input (person_id, start_date, end_date) as
 (
 	select F.person_id, F.start_date, F.end_date
 	FROM (
-	  select I.event_id, I.person_id, I.start_date, E.end_date, dense_rank() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal 
+	  select I.event_id, I.person_id, I.start_date, E.end_date, row_number() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal
```
chrisknoll: @gowthamrao : this is one change that wasn't to revert dense_rank(): I think in this case row_number() is correct: the query is saying 'give me the first end date that happens on or after the qualified event'.   dense_rank() would allow possibly 2 records of the same end_date have ordinal = 1, while row_number() makes sure there's just 1.  dense_rank() vs. row_number() here actaully won't change the era collapsing logic (it will just have a second record to collapse into the same era) but I wanted to make sure I wasn't mising something else.  I think in cases where you want to say 'select exactly one record' you want to use row_number() = 1 where if you want to say 'select all rows that share the earliest date' you would use dense_rank() = 1.

Issue #21, opened 2017-12-21
Impala typecasting issue


wivern:
There are issues with Impala since Impala doesn`t support automatic type casting.
Errors occures with generate cohort queries and primary events queries.

chrisknoll:
Can you provide specific examples?

chrisknoll:
Sorry, I didn't notice the PR you submitted. Will work through that.

Issue #20, opened 2017-12-21
Impala issue with distinct on group by queries


wivern:
Impala doesn`t support distinct with `group by` queries, 
moreover using **distinct** seems redundant with such kind of queries.
Example from src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql:

````sql
select distinct @eraGroup , min(b.start_date) as start_date, b.end_date
	from
		(select distinct @eraGroup, groupid from cteSource) as a
	inner join
		cteEnds as b
	on a.groupid = b.groupid
	group by @eraGroup, end_date
```

chrisknoll:
Can you confirm this is an issue with the current release of circe?  We rewrote those era queries to not use @eraGroup.

wivern:
Oh, sorry. I`ve tested with 1.3.1 release and this error has gone. 
So issue can be closed.

PR #22, opened 2017-12-21
Impala support

wivern:
for #20 and #21 and #23


chrisknoll:
Hi, @wivern,
Thanks for looking into these changes to support Impala.  The eraConstructor query looks good (although we're not directly using that code at the moment), but I do have concern with the CAST(visit_id as int):  visit_occurrence_id should be bigint in the CDM schema. so are you certain the cast to int is valid here?  This will break large datasets with lots of visits...

wivern:
Hi, @chrisknoll ,
You`re absolutely right, I`ve changed CAST to bigint.

Issue #29, opened 2018-02-11
Vocabulary schema is not supported


wivern:
Now it`s limited to utilize the same schema for cdm and vocabularies.

chrisknoll:
That's correct.  The circe expression runs off of a CDM schema (which includes vocabulary tables) not a 'vocabulary schema'.  The notion of a 'vocabulary schema' doesn't actually exist, I think you are thinking about the "vocabulary daimon' in WebAPI where you can have a subset of the CDM tables (ie: the vocabulary tables) and do vocabulary queries against that daimon.  

If you wanted to share vocabulary tables across different CDM schemas, the approach for that is to create views onto your 'root' vocabulary tables so that you don't need to duplicate the data.  I understand your use case for wanting this feature, but it's very important to the circe expression that the patient-level data that is queried is referencing the same vocabulary tables within the same CDM.

pavgra:
@chrisknoll, although we can resolve the limitation by views, what is the problem with making circe-be more flexible in this case and allowing to setup vocab schema differing from the main one? Moreover this would remove the misleading with source daimons in WebAPI.

chrisknoll:
Describing this change as being 'more flexible' is a strawman.  Of course I'm for improving the code by added flexibility, but this feature does 2 things:
1. forces you to pass in a vocabulary schema
2. opens the possibility that  you can now pass an incorrect vocabulary schema that isn't the same vocabulary as the vocabulary used in the CDM you are querying.  Without this feature, you can't make this mistake.

So, if we want to go forward with this to have the 'user beware' attitude where they could mismatch the vocabulary with the CDM and export a set of vocabulary tables where no documentation exists on how to set up this mode, at the very least we should not require the vocabulary schema setting.  The code:

From https://github.com/OHDSI/circe-be/pull/30/files#diff-e6a86866fbd0d23e02d938576ce96f9aR441
```
     if (options.vocabularySchema != null)
       resultSql = StringUtils.replace(resultSql, "@vocabulary_database_schema", options.vocabularySchema);
```

Should do something like this:
```
     if (options.vocabularySchema != null)
       resultSql = StringUtils.replace(resultSql, "@vocabulary_database_schema", options.vocabularySchema);
    else if (options.cdmSchema != null)
       resultSql = StringUtils.replace(resultSql, "@vocabulary_database_schema", options.cdmSchema);
```
That way, existing code that doesn't pass in the parameter will continue to work.

pavgra:
[...]

PR #30, opened 2018-02-11
Atlp 118

wivern:
for #29

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
@@ -438,8 +438,11 @@ public class CohortExpressionQueryBuilder implements IGetCriteriaSqlDispatcher, 
         resultSql = StringUtils.replace(resultSql, "@target_database_schema.@target_cohort_table", options.targetTable);
       if (options.resultSchema != null)
         resultSql = StringUtils.replace(resultSql, "@results_database_schema", options.resultSchema);
-      if (options.vocabularySchema != null)
+      if (options.vocabularySchema != null) {
         resultSql = StringUtils.replace(resultSql, "@vocabulary_database_schema", options.vocabularySchema);
+      } else if (options.cdmSchema != null) {
+        resultSql = StringUtils.replace(resultSql, "@vocabulary_database_schema", options.cdmSchema);
+      }
       if (options.cohortId != null)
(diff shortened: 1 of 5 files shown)


chrisknoll:
Hi, Guys. At the risk of 'santa-barbara'-ing this request:  We've asked that the odysseusinc to please allow us to talk about a proposed change before you guys go ahead and implement. What happens is that there's some reason why something works the way it works (whether it be a vocabulary schema, relative paths in module_ids in the atlas charts library, or how the results DDL should be structured in the codebase) and when we have some corrections, you guys probably feel like we're introducing drama into the process.  

Therefore, please submit your issue, let us discuss the details, and then proceed to implement.

Re: on this PR, I've made comments over in the issue thread about the reason why we don't have a vocabulary schema property.  If there's other people that want to chime in on this such as @anthonysena , @jduke99 , @cgreich , or others, we can debate if we should enforce that the CDM should contain the vocabulary tables or if a CDM can be built without those vocabulary tables.  If we want to treat those things as separate schemas then we can change circe-be to support that. But for now, there's only a CDM schema.

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
```diff
@@ -435,6 +438,8 @@ public String buildExpressionQuery(CohortExpression expression, BuildExpressionQ
         resultSql = StringUtils.replace(resultSql, "@target_database_schema.@target_cohort_table", options.targetTable);
       if (options.resultSchema != null)
         resultSql = StringUtils.replace(resultSql, "@results_database_schema", options.resultSchema);
+      if (options.vocabularySchema != null)
```
chrisknoll: This should be changed so that the vocabulary schema is optional.  If you have a CDM schema and the vocabulary shema is null, then replace vocabulary schema with the cdmSchema value.

chrisknoll:
@pavgra , @wivern , I know I said that we want to discuss this with the CDM folks about vocabulary and patient-level data separation, however, I also wanted to get the 1.4 release out there. So, how about you make the changes to make the vocabulary schema optional, I'll make the merge for circe-be 1.4, but we really should establish the recommended approach when dealing with 'shared' vocabularies.  I can always remove this configuration option later if it is deemed inappropriate, but for now, for the sake of closing out this release, if you can make the requested changes, I'll merge this in and it will be apart of WebAPI 2.3.

PR #32, opened 2018-02-27
Null cast issue

pavgra:
For https://github.com/OHDSI/circe-be/issues/31

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>org.ohdsi</groupId>
     <artifactId>circe</artifactId>
-    <version>1.4.0-SNAPSHOT</version>
+    <version>1.4.1-SNAPSHOT</version>
 
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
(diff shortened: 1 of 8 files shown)


chrisknoll:
Ok, I'm going to make an unorthodox maneuver where I'm going to re-issue the 1.4.0 release with this fix included.

Thanks for the fast turnaround, @pavgra !

pavgra:
@chrisknoll, but is it a problem to change version in WebAPI since the release is not done yet? Just would be great to keep versioning and artifacts repo clean and not confusing

chrisknoll:
@pavgra , probably, but I did a 'bad thing' because I didn't want to interrupt the releases going on in WebAPI (they've already set their dependencies up to point to 1.4.0 and I didn't want to bother them again with another update).  So, in the future, I'll just push out another point-release (1.4.1 in this case) but for this one-time-only bad behavior, I've upated the commit to point to set the pom.xml to 1.4.0, and re-issued the 1.4.0 release.  I know, it's very dirty, I hope you can still respect me inspite of this weakness.

PR #36, opened 2018-04-06
Fix feature extraction where censor period is being used

wivern:
fixes OHDSI/WebAPI#364


chrisknoll:
Hi, we're talking about the implications of this change in the issue for WebAPI: OHDSI/WebAPI#364.  Once we figure out what the right path is, we'll revisit this PR.

chrisknoll:
I believe this PR is obsolete, since we've now added an inclusion report at the Person level, and censoring window doesn't impact the selection of inclusion events (just truncates/censors the resulting cohort episodes).

I think there is a need to record number of people censored, but it won't be accomplished int he way this PR has materialized, so if you want to work on the question of 'how many people censored?', let's start with a fresh branch from master and determine how best to achieve this goal.

PR #37, opened 2018-04-16
Provide diagnostic checks for cohorts

wivern:
closes OHDSI/Atlas#498

### DIFF
--- a/src/main/java/org/ohdsi/circe/check/Constants.java
+++ b/src/main/java/org/ohdsi/circe/check/Constants.java
@@ -1,0 +1,72 @@
+/*
+ *   Copyright 2017 Observational Health Data Sciences and Informatics
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ *
+ *   Authors: Vitaly Koulakov
+ *
+ */
+
+package org.ohdsi.circe.check;
+
+public interface Constants {
+
+    interface Criteria {
+
+        String CONDITION_ERA = "condition era";
+        String CONDITION_OCCURRENCE = "condition occurrence";
+        String DEATH = "death";
+        String DEVICE_EXPOSURE = "device exposure";
+        String DOSE_ERA = "dose era";
+        String DRUG_ERA = "drug era";
+        String DRUG_EXPOSURE = "drug exposure";
+        String MEASUREMENT = "measurement";
+        String OBSERVATION = "observation";
+        String PROCEDURE_OCCURRENCE = "procedure occurrence";
+        String SPECIMEN = "specimen";
+        String VISIT_OCCURRENCE = "visit occurrence";
+        String PAYER_PLAN_PERIOD = "payer plan period";
+        String OBSERVATION_PERIOD = "observation period";
+    }
+
+    interface Attributes {
+
+        String AGE_ATTR = "age";
+        String QUANTITY_ATTR = "quantity";
+        String OCCURRENCE_START_DATE_ATTR = "occurrence start date";
+        String OCCURRENCE_END_DATE_ATTR = "occurrence end date";
+        String ERA_START_DATE_ATTR = "era start date";
+        String ERA_END_DATE_ATTR = "era end date";
+        String DOSE_VALUE_ATTR = "dose value";
+        String ERA_LENGTH_ATTR = "era length";
+        String AGE_AT_START_ATTR = "age at start";
+        String AGE_AT_END_ATTR = "age at end";
+        String OCCURRENCE_COUNT_ATTR = "occurrence count";
+        String GAP_DAYS_ATTR = "gap days";
+        String AGE_AT_ERA_START_ATTR = "age at era start";
+        String AGE_AT_ERA_END_ATTR = "age at era end";
+        String REFILLS_ATTR = "refills";
+        String DAYS_SUPPLY_ATTR = "days supply";
+        String EFFECTIVE_DRUG_DOSE_ATTR = "effective drug dose";
+        String VALUE_AS_NUMBER_ATTR = "value as number";
+        String RANGE_LOW_ATTR = "range low";
+        String RANGE_HIGH_ATTR = "range high";
+        String RANGE_LOW_RATIO_ATTR = "range low ratio";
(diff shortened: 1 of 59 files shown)


src/main/java/org/ohdsi/circe/check/Constants.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ */
+
+package org.ohdsi.circe.check;
+
+public interface Constants {
+
+    interface Criteria {
+
+        String CONDITION_ERA = "condition era";
```
chrisknoll: Why not enums?

chrisknoll:
Hi, I've done a little testing, and I have a few questions:

See http://www.ohdsi.org/web/atlas/#/cohortdefinition/1769335 for reference cohort definitoin.

When I generate doagonstics, I get 40 warnings.  I'll go through the different 'types' and understand if this is correct:

```
Concept Set "Long-term care visit" is not usedFix It
--

Works!  clicking fix it removes the concept set from the cohort. Nice.

```

Provider specialty is empty at inclusion rule Prior Atrial Fibrillation in condition occurrence criteria.
 Provider specialty is empty at inclusion rule No Heart Valve or mitral stenosis in condition occurrence criteria.
Provider specialty is empty at inclusion rule No Kidney Procedures in condition occurrence criteria.
Provider specialty is empty at inclusion rule No hemodialysis in condition occurrence criteria.
```
Provider specialty is not specified in any of these inclusion rules, and i'm not sure if this is only happening for specific criteria types (ie: condition) or what, but there are 8 inclusion rules, but this warning only came up 4 times. So, I don' think any of these should have appeared at all.

```

drug exposure criteria at inclusion rule No Anticoagulants time window breaks common pattern
--
procedure occurrence criteria at inclusion rule No joint replacement time window breaks common pattern
observation criteria at inclusion rule No hemodialysis time window breaks common pattern
condition occurrence criteria at inclusion rule No hemodialysis time window breaks common pattern
procedure occurrence criteria at inclusion rule No hemodialysis time window breaks common pattern
```

Not sure what 'pattern' is being broken, but these rules say 'at most 0 of X between 183 d prior and 0d prior', shouldn't that be a valid pattern?

```
It's not specified what type of records to look for in a drug exposure at initial event
```
I think this is a fair warning that people should know that they could specify domain_type_concepts in their criteria, but normally, people do not.  And, if that's intended, there's no way to 'fix' this message.  Maybe this could be specified as an 'info' class of warning.  Also, it might be helpful to just capture that one warning globally for the cohort definition instead of repeating it for each time you found it.

```
condition occurrence at inclusion rule Prior Atrial Fibrillation didn't specify that it must be first time in patient's history
```

This one also repeats for all criteria in the expression.   Note, it's only a 'information' level warning, but I don't think we should tell people that they should use 'first' as a matter of diagnostics.  There are different types of checks that we could do on 'firsts' such as if someone tries to count > 1 events that are said to 'use first', but that's a pretty edge use case, and I'm not sure I would worry about it.

Anyways, if you would like to copy the cohort definition from the link above into your local environment, you should see the errors and warnings.  

Also, I'm wondering if a filter function on the warning tab to filter by severity so by default you see errors and warnings, but info can be suppressed by default.

**Comment by wivern** [comment]
Hi, Chris! 
Thank you for this great job on diagnostics testing,
i'll double check the results and implement fixes.

**Comment by wivern** [comment]
1. Provider specialty
It follows the rule:
Condition Occurrence Criteria: Provider Speciality is empty (1 – Error)

I guess this rule is specific for Condition Occurrence Criteria only. 	Is this rule actual or it requires some corrections?

2. Time window pattern
Rule:
"Many times your review of time before and after index in the inclusion rules are exactly the same (e.g. -365,0).  If one of your rules breaks the pattern, warn the user."

Fixed message, now it would be “drug exposure criteria at inclusion rule No Anticoagulants time window differs from most common pattern prior 'all days before and 0 days before ', shouldn't that be a valid pattern?”

3. not specified what type of records to look for

Collapsed to one message and it`s severity reduced to Info.

4.  "If requiring a clean period (e.g., 365 days pre-index without having the exposure), but didn’t specify that it must be first time in patient’s history, have a check that this is OK or if the user wants instead to use first time in patient history."

This rule was turned off.

**Comment by chrisknoll** [comment]
For #1: it's not a required parameter to specify a Provider. So if it is empty, it's not an error.  It's not even a WARN or INFO...it's valid to create a criteria with no provider specialty  selected.

**Comment by chrisknoll** [comment]
@wivern , I just pushed a quick change to check the 'limit progression' check properly:  the query builder will ignore the qualifiedLimit if additionalCriteria is empty (null).  I was getting warnings about the progression on a cohort def that didn't have additionalCriteria specified, so that's why I changed it.

I was going to make a couple other changes, but I think there's a basic misunderstanding about what I said earlier, and wanted to clarify it with you:

The DomainType check that we have is a general Info message where people are notified that they are not specifying the type of DomainType to look for in the record. I said this was a fair warning, but there's another type of check that should be done that is more of a 'warning' class (we can call this check 'EmptyDomainTypeCheck') that if the type criteria is not null && type is empty, then we should warn that they did not specify a type concept list in the criteria.

Ie: here:
```
                        .when(conditionOccurrence -> Objects.isNull(conditionOccurrence.conditionType) || conditionOccurrence.conditionType.length == 0)
                        .then(addWarning);
```
Btw: i love the way you applied lambda expressions and Operators here to apply these warnings. that's very cool.

This check works for just giving the info message that they should think about specifying a domain type for these criteria.  HOwever, there's a Warning check that shoudl work like the genderConcept check:

```
                        .when(conditionOccurrence -> Objects.isNull(conditionOccurrence.conditionType) && conditionOccurrence.conditionType.length == 0)
                        .then(addWarning))
```
Ie: It's a warning when they asked to specify a domain type but they left the domain type empty.

So, I think all that's required here is to add a new DomainTypeMissingCheck, which will trigger for a criteria when things like conditionType is not null but .length == 0.

wivern:
@chrisknoll thank you for explanations. It was really some misunderstood since the application area still is not pretty clean to me.
I've added new EmptyDomainTypeCheck rule and now it seems DomainTypeCheck sould warn only if domain type was not set because case when domain type is empty covered by EmptyDomainTypeCheck now, right?

chrisknoll:
@wivern : that's right:  DominTypeCheck should warn ('info' severity) if it's not set (ie, it is null) so that we can give people information that it's a good idea to specify the domain type (ie: patient reported vs. physician diagnosed).  But it's not required, and not an error, hence 'info' severity.

Separately, we have a EmptyCheck ('warn' severity) that if a person _does_ specify a domainType criteria (ie: a Concept[]), but the array is empty, we should warn that they left the domainType criterion empty.

chrisknoll:
Ok, so first thing: @wivern : good job on this work, I really like the techniques that you applied here to make the check-logic readable.  

I'm inclined to take this PR into master, even tho I don't think I've given it enough depth of review and testing.  I think that I'd like to get a broader audience to start using these checks in their real-world work so that we can see if there's any logical problems with these checks or gaps based on real-world circumstances.

So, I'll merge this into master now.  The merges for the Atlas UI changes are on hold until the refactoring changes are made, at which point OHDSI/Atlas#647 will probably need to be re-based/merged to get the refactoring changes incorporated.

PR #41, opened 2018-06-28
Fixed error in impala

agackovka:
closes https://github.com/OHDSI/WebAPI/issues/457

### DIFF
--- a/src/main/resources/resources/cohortdefinition/sql/customEraStrategy.sql
+++ b/src/main/resources/resources/cohortdefinition/sql/customEraStrategy.sql
@@ -3,11 +3,9 @@
 select de.PERSON_ID, DRUG_EXPOSURE_START_DATE,  COALESCE(DRUG_EXPOSURE_END_DATE, DATEADD(day,DAYS_SUPPLY,DRUG_EXPOSURE_START_DATE), DATEADD(day,1,DRUG_EXPOSURE_START_DATE)) as DRUG_EXPOSURE_END_DATE 
 INTO #drugTarget
 FROM @cdm_database_schema.DRUG_EXPOSURE de
-  WHERE EXISTS (SELECT 1 FROM @cdm_database_schema.DRUG_EXPOSURE dei
-    JOIN @eventTable events on events.person_id = de.person_id
-    JOIN #Codesets cs on cs.codeset_id = @drugCodesetId AND
-      (de.drug_concept_id = cs.concept_id OR de.drug_source_concept_id = cs.concept_id)
-    )
+JOIN #Codesets cs on cs.codeset_id = @drugCodesetId 
+	AND (de.drug_concept_id = cs.concept_id OR de.drug_source_concept_id = cs.concept_id)
+WHERE de.person_id in (select person_id from @eventTable)
 ;
 
 select et.event_id, et.person_id, ERAS.era_end_date as end_date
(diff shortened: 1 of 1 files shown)


src/main/resources/resources/cohortdefinition/sql/customEraStrategy.sql
```diff
@@ -3,9 +3,9 @@
 select de.PERSON_ID, DRUG_EXPOSURE_START_DATE,  COALESCE(DRUG_EXPOSURE_END_DATE, DATEADD(day,DAYS_SUPPLY,DRUG_EXPOSURE_START_DATE), DATEADD(day,1,DRUG_EXPOSURE_START_DATE)) as DRUG_EXPOSURE_END_DATE 
 INTO #drugTarget
 FROM @cdm_database_schema.DRUG_EXPOSURE de
-WHERE (de.drug_concept_id in (SELECT concept_id from  #Codesets where codeset_id = @drugCodesetId)
```
chrisknoll: This change isn't yielding the same sub query results:  the logic is to only find drug exposures for people in the cohort (hence the ```where person_id in (select person id from @eventTable)```) .  The switch to using joins means that you get multiple rows per person in the event table in the result, and I'm not sure what impact that will have to the overall query...but it's not the same.

chrisknoll: It's possible (but I am still unhappy if we have to write query logic to cater to one specific cialect) that the above WHERE (A OR B) and C caould be rewritten as NOT (NOT A AND NOT B) AND C so the sql would be:

```
WHERE NOT (
  de.drug_concept_id NOT in (SELECT concept_id from  #Codesets where codeset_id = @drugCodesetId) 
  AND de.drug_source_concept_id NOT in (SELECT concept_id from  #Codesets where codeset_id = @drugCodesetId)
)  AND de.person_id in (select person_id from @eventTable)
```

agackovka: @chrisknoll maybe it's better to write something like this:
```FROM @cdm_database_schema.DRUG_EXPOSURE de
  WHERE de.drug_exposure_id IN (
    SELECT dei.drug_exposure_id FROM @cdm_database_schema.DRUG_EXPOSURE dei
      JOIN @eventTable events on events.person_id = de.person_id
      JOIN #Codesets cs on cs.codeset_id = @drugCodesetId AND
        (de.drug_concept_id = cs.concept_id OR de.drug_source_concept_id = cs.concept_id)
)
;```
i think it's more 'natural' and hope works same

Comment (chrisknoll) [reply src/main/resources/resources/cohortdefinition/sql/customEraStrategy.sql]: Does ```exists``` work in Impala
```
where exists (select 1 from FROM @cdm_database_schema.DRUG_EXPOSURE dei
      JOIN @eventTable events on events.person_id = de.person_id
      JOIN #Codesets cs on cs.codeset_id = @drugCodesetId AND
        (de.drug_concept_id = cs.concept_id OR de.drug_source_concept_id = cs.concept_id)
)
```

**Comment by agackovka** [comment]
@chrisknoll updated

**Comment by chrisknoll** [comment]
hey @agackovka , I'm sorry, but when I tried to run with the latest changes, the query doesn't return after 30 mintues (when my test completes in under 3 minutes).

I'm looking at the exists, and I think there's a problem with the subquery (and I apologize I gave you this suggestion, but I don't see how it would work on impala):

```
where exists (select 1 from FROM @cdm_database_schema.DRUG_EXPOSURE dei
      JOIN @eventTable events on events.person_id = de.person_id
      JOIN #Codesets cs on cs.codeset_id = @drugCodesetId AND
        (de.drug_concept_id = cs.concept_id OR de.drug_source_concept_id = cs.concept_id)
)
```

The problem with the above is that the 'exists' statementn jonis to DRUG_EXPOSURE dei, but never joins the dei to event table...and so I'm wondering if this is some kind of cartesian product.

I'm trying to work through the logic of the query in my mind, but basically we want to get all the drug exposure records for the people in the eventTable and also matching the drug_concept_id in #codesets in the drug_concept_id or the drug_source_concept_id table.

maybe exists isn't the way here.. to be honest, the origional form of the query was the most logcially concise...and I'm sorry Impala can't handle the OR statement...but...what we need this query to do is:

Return drug_exposure records where the drug_expsoure is for a person in @eventTable, and the drug_concept_id or drug_source_concept id are found in #codesets.

Can you come up with an impala compatable query that does this logic?

chrisknoll:
@agackovka , can you pull this branch down and confirm on Impala? It runs fine on PDW.

pavgra:
@chrisknoll, I've tested - the error in Impala has gone

PR #42, opened 2018-07-09
Domain type exclusions and query optimizations.

### LINKED ISSUE #40: Support domain type criteria for NOT IN()
Currently, if you want to exclude patient reported records, you would have to filter based on the record having any concept type EXCEPT the concept you want to exclude from finding.

This request is to support a 'domain type EXCEPT' criteria where you an pass in a set of Type Concepts but will return rows that are NOT IN the specified list.

chrisknoll:
Allow type concepts to be used as exclusions.
Optimization: apply left join only when looking for counts that might be 0.
Usability: do not create subquery for empty criteria groups.

Fixes #40.

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
@@ -642,7 +642,7 @@ public class CohortExpressionQueryBuilder implements IGetCriteriaSqlDispatcher, 
     );
 		
 		// join type is LEFT when counts of 0 or 'at most' is specified
-		String joinType = (corelatedCriteria.occurrence.type == 1 || corelatedCriteria.occurrence.count == 0) ?  "LEFT" : "INNER";
+		String joinType = (corelatedCriteria.occurrence.type == Occurrence.AT_MOST || corelatedCriteria.occurrence.count == 0) ?  "LEFT" : "INNER";
     
     query = StringUtils.replace(query, "@joinType", joinType);
 
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
@@ -486,20 +486,20 @@ public class CohortExpressionQueryBuilder implements IGetCriteriaSqlDispatcher, 
       if (group.type.equalsIgnoreCase("ANY")) // count must be > 0 for an 'ANY' criteria
         occurrenceCountClause += "> 0"; 
 
-      if (group.type.toUpperCase().startsWith("AT_"))
-      {
-        if (group.type.toUpperCase().endsWith("LEAST"))
-          occurrenceCountClause += ">= " + group.count;
-				else {
-          occurrenceCountClause += "<= " + group.count;
+			if (group.type.toUpperCase().startsWith("AT_")) {
+				if (group.type.toUpperCase().endsWith("LEAST")) { // AT_LEAST
+					occurrenceCountClause += ">= " + group.count;
+				} else { // AT_MOST, which includes zero
+					occurrenceCountClause += "<= " + group.count;
 					joinType = "LEFT";
 				}
-				
-				if (group.count == 0) {
-					joinType = "LEFT"; //if you are looking for a zero count, you need to do a left join
+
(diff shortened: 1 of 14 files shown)


src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
```diff
@@ -634,8 +640,13 @@ public String getCorelatedlCriteriaQuery(CorelatedCriteria corelatedCriteria, St
       getOccurrenceOperator(corelatedCriteria.occurrence.type), 
       corelatedCriteria.occurrence.count
     );
+		
+		// join type is LEFT when counts of 0 or 'at most' is specified
+		String joinType = (corelatedCriteria.occurrence.type == 1 || corelatedCriteria.occurrence.count == 0) ?  "LEFT" : "INNER";
```
pavgra: Where is the type value (1) coming from? - would you mind to move it into constants (or re-use existing one)?

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
```diff
@@ -509,8 +515,8 @@ public String getCriteriaGroupQuery(CriteriaGroup group, String eventTable) {
   private String getInclusionRuleQuery(CriteriaGroup inclusionRule)
   {
     String resultSql = INCLUSION_RULE_QUERY_TEMPLATE;
-    String additionalCriteriaQuery = "\nJOIN (\n" + getCriteriaGroupQuery(inclusionRule, "#qualified_events") + ") AC on AC.person_id = pe.person_id AND AC.event_id = pe.event_id";
-    additionalCriteriaQuery = StringUtils.replace(additionalCriteriaQuery,"@indexId", "" + 0);
+		String additionalCriteriaQuery = "\nJOIN (\n" + getCriteriaGroupQuery(inclusionRule, "#qualified_events") + ") AC on AC.person_id = pe.person_id AND AC.event_id = pe.event_id";
```
pavgra: Diff is effectively empty and annotations in git history will be irrelevant

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
```diff
@@ -490,15 +490,21 @@ public String getCriteriaGroupQuery(CriteriaGroup group, String eventTable) {
       {
         if (group.type.toUpperCase().endsWith("LEAST"))
           occurrenceCountClause += ">= " + group.count;
-        else
+				else {
```
pavgra: I'd suggest to either use brackets in all parts of `if` or in none - from a codestyle perspective and easy-to-read perspective.

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
```diff
@@ -490,15 +490,21 @@ public String getCriteriaGroupQuery(CriteriaGroup group, String eventTable) {
       {
         if (group.type.toUpperCase().endsWith("LEAST"))
```
pavgra: Wouldn't strategy pattern fit here better? - now it's hard to read this nested `if`

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
chrisknoll: Ok, will do.

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
chrisknoll: I'll revert these changes to the space-indents.  This was a result of trying somethign and reversing it, but the reversal left the tab-indents.

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
chrisknoll: Actually, need to correct this:  you should only check for a group.count inside the AT_LEAST/AT_MOST if block, because group count could be zero (by default) but the group.type may be ALL or ANY which should result in a INNER JOIN, not a LEFT join. So, there will be 2 if statemetns inside the "AT_" check. I don't think it's hard to read tho.

PR #49, opened 2018-07-30
By Person report for Inclusion Rule Impact Analysis.

### LINKED ISSUE #47: Inclusion Rule reports: report at person level
### Current Behavior
The inclusion rule report evaluates all qualified events (those events selected via 'initial events' and then filtered by 'initial event criteria') which gives you an 'event-level' report for the inclusion criteria.  If the cohort was defined as using 'one event per person' (ie: earliest/latest), then you can consider the report 'person-centric'. However, there are many more use cases of late that would allow a person to appear multiple times in a cohort, so the current reporting behavior at the event-level is not effective for understanding 'distinct people'.

### Proposed Change
In order to get the reports to show 'person-centric' data, we need to limit the events in the report to one-per-person.  One approach to this event selection would be to chose the event that had the most number of inclusion rules satisfied.  This event would represent the 'best case' for a person to match the most of the inclusion rules.  

### Considerations

Handling 'ties'.  Take this example: there are 3 inclusion rules, and a person satisfies 1,2 at e1 and 2,3 at e2  From the proposed solution, those events are 'tied' at 2 inclusion rules.  To break this tie, we'll select the event that has the lowest inclusion rule satisfied. In this case, we'll select the event that satisfies rules 1,2.

Reporting the 'best' event: if person satisfies 1,2 at e1, but 2,3,4 at e2. e2 will be selected for reporting even tho e1 matched a lower inclusion rule.  The final report will not report this person matched inclusion rule 1 because the event selected only matched 2,3,4.  Otherwise, we would be double-counting people and our reports would no longer be person-centric.

chrisknoll:
Fixes #47

### DIFF
--- a/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
+++ b/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
@@ -142,95 +142,14 @@ WHERE ranked.rank_value = 1
 -- 0: all events
 -- 1: best event
 
--- calculte matching group counts - event
-delete from @results_database_schema.cohort_inclusion_result where cohort_definition_id = @target_cohort_id and mode_id = 0;
-insert into @results_database_schema.cohort_inclusion_result (cohort_definition_id, inclusion_rule_mask, person_count, mode_id)
-select @target_cohort_id as cohort_definition_id, inclusion_rule_mask, count(*) as person_count, 0 as mode_id
-from
-(
-  select Q.person_id, Q.event_id, CAST(SUM(coalesce(POWER(cast(2 as bigint), I.inclusion_rule_id), 0)) AS bigint) as inclusion_rule_mask
-  from #qualified_events Q
-  LEFT JOIN #inclusion_events I on q.person_id = i.person_id and q.event_id = i.event_id
-  GROUP BY Q.person_id, Q.event_id
-) MG -- matching groups
-group by inclusion_rule_mask
-;
 
--- calculte matching group counts - person
-delete from @results_database_schema.cohort_inclusion_result where cohort_definition_id = @target_cohort_id and mode_id = 1;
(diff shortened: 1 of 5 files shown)


src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
```diff
@@ -147,22 +184,57 @@ left join
...
 CROSS JOIN (select count(event_id) as total from #qualified_events) EventTotal
-LEFT JOIN @results_database_schema.cohort_inclusion_result SR on SR.cohort_definition_id = @target_cohort_id AND (POWER(cast(2 as bigint),RuleTotal.total_rules) - POWER(cast(2 as bigint),ir.rule_sequence) - 1) = SR.inclusion_rule_mask -- POWER(2,rule count) - POWER(2,rule sequence) - 1 is the mask for 'all except this rule' 
+LEFT JOIN @results_database_schema.cohort_inclusion_result SR on SR.mode_id = 0 AND SR.cohort_definition_id = @target_cohort_id AND (POWER(cast(2 as bigint),RuleTotal.total_rules) - POWER(cast(2 as bigint),ir.rule_sequence) - 1) = SR.inclusion_rule_mask -- POWER(2,rule count) - POWER(2,rule sequence) - 1 is the mask for 'all except this rule' 
+WHERE ir.cohort_definition_id = @target_cohort_id
+;
+
+-- calculate gain counts - person
+delete from @results_database_schema.cohort_inclusion_stats where cohort_definition_id = @target_cohort_id and mode_id = 1;
+insert into @results_database_schema.cohort_inclusion_stats (cohort_definition_id, rule_sequence, person_count, gain_count, person_total, mode_id)
```
pavgra: I see that the SQL blocks for event and person level stats are very similar, but I understand that you want them separate from logical perspective. What if we move the statements to separate sub-templates and insert it into the generateCohort.sql file? Then we'll remove duplication from the codebase

PR #56, opened 2018-09-26
Adding fromJson to OutcomeCohortExpression; alignment to specifications

anthonysena:
Updates for serializing negative control outcome expressions from JSON and aligning the property names to the specification.

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/negativecontrols/OutcomeCohortExpression.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/negativecontrols/OutcomeCohortExpression.java
@@ -2,6 +2,7 @@ package org.ohdsi.circe.cohortdefinition.negativecontrols;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/org/ohdsi/circe/cohortdefinition/negativecontrols/OutcomeCohortExpression.java
```diff
@@ -16,17 +17,16 @@
...
-		try {
-			CohortExpression expression = JSON_MAPPER.readValue(json, CohortExpression.class);
-			return expression;
-		} catch (Exception e) {
-			throw new RuntimeException("Error parsing cohort expression", e);
-		}
-	}
- */
+ public static OutcomeCohortExpression fromJson(String json) {
+   try {
```
pavgra: https://github.com/OHDSI/StandardizedAnalysisAPI/blob/pathways-analysis/src/main/java/org/ohdsi/analysis/Utils.java#L67

anthonysena: That makes sense to me @pavgra; however, I'm not sure if we're leveraging SAA in circe just yet? I'll leave it to @chrisknoll's discretion on how to proceed.

chrisknoll: Post Symposium.

anthonysena: To make sure I'm clear: we can refactor to use the SAA Utils for formatting after symposium. Can we merge this change in with that understanding? I see @pavgra linked it to the WebAPI issue. I need this functionality for Symposium so if we can merge it in, that would be greatly appreciated. @schuemie needs this for testing on Hydra.

chrisknoll: Ok, it will be done.

PR #57, opened 2018-09-27
Modified CriteriaGroup JSON Seralization

anthonysena:
When serializing a CohortExpression, I found that the CriteriaGroup was inserting an extraneous element:

```
"AdditionalCriteria": {
       "empty": true,
       "Type": "ANY",
       "Count": null,
       "CriteriaList": [],
       "DemographicCriteriaList": [],
       "Groups": []
},

```

Note the `"empty":true` property above. This was causing issues when attempting to use `CohortExpression.fromJson()` to re-hydrate the definition. I modified the CriteriaGroup to ignore the `isEmpty` function in the class and that fixed the issue.

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CriteriaGroup.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CriteriaGroup.java
@@ -18,6 +18,7 @@
  */
 package org.ohdsi.circe.cohortdefinition;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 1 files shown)


pavgra:
@anthonysena / @chrisknoll , this is a workaround. Your serialization function is incorrect, not the DTO. Pls get a look at how serialization is done in SAA:
https://github.com/OHDSI/StandardizedAnalysisAPI/blob/master/src/main/java/org/ohdsi/analysis/Utils.java#L18

PR #61, opened 2018-10-30
separates cohort expression inserts and cleanup

wivern:
this improvement is required to https://github.com/OHDSI/WebAPI/issues/495#issuecomment-412667979

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
@@ -21,6 +21,7 @@ package org.ohdsi.circe.cohortdefinition;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.ArrayList;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.ohdsi.circe.helper.ResourceHelper;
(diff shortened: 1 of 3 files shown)


chrisknoll:
Could you go into more detail what the requirement is?

I had thought about splitting up the query into pieces so that you could 'inject' custom steps into the cohort generation, but in the end, since you can define a 'target cohort table' in the process, we'd simply create a cohort table up front (even if it is temporary), generate the cohort into the temp table, then do any post-process work on the target cohort table:

```
1. create target cohort table
2. run generate cohort specifying custom target cohort table (this will perform cohort generation specific cleanup)
3. Use data from target cohort table for work.
4. cleanup target cohort table (if necessary).
```

Could you explain your use-case?

chrisknoll:
There's a mix of tabs and spaces for indentation in this file, and this is most likely due to the legacy nature of some of these source files. However, tabs is the specified indentation, so any code you add or update, please re-format to single tab indent.

wivern:
@chrisknoll The use-case you described is exactly the same as we implement for cohort characterization precedence analysis. But when applying distribution analysis we need to access to internal calculations of events. It`s not possible when cohort generation process is solid since temporary events tables dropped after cohort generation has finished.

chrisknoll:
@wivern,  thanks for cleaning up the whitespace. what I meant in my comment, tho, was that you just needed to clean up the whitespace in the specific lines you touched, not reformat the entire file. but, since it's done now, don't worry about it. just for future reference: if we run across a file with the old indentation style, just apply new indentation style to the lines you are changing.  Eventually we will want to just reformat all the files to the tabs-indentation in 1 commit.

>  But when applying distribution analysis we need to access to internal calculations of events. 

I'm meeting with @pavgra  today on this.  There is a flaw in the logic that in order to do distribution characterization, we need to break open the cohort definition.  We might want to apply characterization process on a cohort that didn't come from a circe expression, so the process of identifying the records to produce the distribution for should not depend on the method of cohort generation.  @gowthamrao , i believe your influence led to this approach that @wivern / @pavgra is taking here, and we need to discuss this before circe cohort definitions gets warped out of it's intended design: to define cohorts.

pavgra:
@chrisknoll , keeping in mind our today's discussion, I still think that this PR is valuable, and don't see a reason for not merging it. What do you think?

chrisknoll:
@pavgra , I'm not sure:  I don't think splitting the buildQuery into 2 parts makes it more 'atomic', and I consider the circe library _extremely_ well structured.  

But, I think I understand what you are asking for: you want 'hooks' to override certain 'phases' of cohort generation.  Maybe to add these hooks we have specific methods to the query builder for the various phases:  init -> buildCodesetsQuery -> buildCohortEntryEventQuery -> buildInclusionRuleQuery-> buildCohortExitQuery -> buildCohortErasQuery -> cleanup

Something like this? In this way, a consumer of the class can override the method in question to add additional work, and override cleanup if it the new work involves creating new temp tables, for example.

I'm 100% behind adding useful features to the library, I'm just not sure this particular approach really advances the library (no offense intended).

PR #66, opened 2019-01-18
Fix inclusion impact analysis where censor period is being used

wivern:
Fixes OHDSI/WebAPI#364

### DIFF
--- a/src/main/resources/resources/cohortdefinition/sql/cohortInclusionAnalysis.sql
+++ b/src/main/resources/resources/cohortdefinition/sql/cohortInclusionAnalysis.sql
@@ -32,15 +32,21 @@ WHERE ir.cohort_definition_id = @target_cohort_id
 
 -- calculate totals
 delete from @results_database_schema.cohort_summary_stats where cohort_definition_id = @target_cohort_id and mode_id = @inclusionImpactMode;
-insert into @results_database_schema.cohort_summary_stats (cohort_definition_id, base_count, final_count, mode_id, lost_count)
-select @target_cohort_id as cohort_definition_id, PC.total as person_count, coalesce(FC.total, 0) as final_count, @inclusionImpactMode as mode_id, coalesce(FCC.total_people - TC.total, 0) as lost_count
+insert into @results_database_schema.cohort_summary_stats (cohort_definition_id, base_count, final_count, mode_id)
+select @target_cohort_id as cohort_definition_id, PC.total as person_count, coalesce(FC.total, 0) as final_count, @inclusionImpactMode as mode_id
 FROM
 (select count_big(event_id) as total from @eventTable) PC,
-(select count_big(distinct person_id) as total_people from #final_cohort) FCC,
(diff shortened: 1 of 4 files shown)


chrisknoll:
My concern with this is that this is adding a statistic related to censoring (which is applied after all the cohort eras are constructed), while the statistics in this table are about the cohort entry events. 

You are correctly puling in 2 statistics: distinct people in `#final_cohort` vs. distinct people inserted into the cohort table, so subtracting those 2 values does give you the difference between "all people" and "censored people", but putting those statistics along side the 'event-level' statistics that are cohort entry events I am 100% positive will cause confusion for people.

It is a good thing that you did optionally hide the censoring statistics in the front end.  I think it's important that the label for that statistic in the UI is very explicit that those numbers represents people, not events.  

I don't have a proposal here to adjust how the data is stored; what you've provided here I think will work technically. I just wanted to call out here (for any interested readers) that the data stored in cohort_summary_stats (where mode_id = 0 is all events and mode_id 1 is best event) is now a mix of event-level data and person-level data.  If I had to make a proposal, I think I'd introduce a new table called cohort_censor_stats that doesn't have a mode_id in it (because censoring doesn't have a 'best event vs. all events' context, and use this separate table to capture statistics on censoring, isntead of mixing it with he other statistics.

How do you feel about making that sort of change?

src/main/resources/resources/cohortdefinition/sql/cohortInclusionAnalysis.sql
```diff
@@ -32,15 +32,21 @@ WHERE ir.cohort_definition_id = @target_cohort_id
...
 (select count_big(event_id) as total from @eventTable) PC,
-(select count_big(distinct person_id) as total_people from #final_cohort) FCC,
-(select count_big(distinct subject_id) as total from @target_database_schema.@target_cohort_table t where t.cohort_definition_id = @target_cohort_id) TC,
 (select sum(sr.person_count) as total
   from @results_database_schema.cohort_inclusion_result sr
   CROSS JOIN (select count(*) as total_rules from @results_database_schema.cohort_inclusion where cohort_definition_id = @target_cohort_id) RuleTotal
   where sr.mode_id = @inclusionImpactMode and sr.cohort_definition_id = @target_cohort_id and sr.inclusion_rule_mask = POWER(cast(2 as bigint),RuleTotal.total_rules)-1
 ) FC
 ;
+
```
chrisknoll: Put the --- calculate censored into a separate block of SQL because the cohortInclusionAnalysis.sql is called twice, once for mode 0 and mode 1, while censor stats is calcuated once. 

Put another way:  censor stats isn't part of 'cohrotInclusionAnalysis' so it shoudln't be contained in 'cohortInclusionAnalysis.sql'.

chrisknoll:
I just had to make an adjustment to the censor statistics calculation:  you were checkign censor  criteria which is different from the censor window: censor criteria are used to abort a cohort event before the event's end date based on identifying some record in the data via criteria.  

**Censor Window**, on the other hand, is used to take the set of cohort eras, and left/right trim them to a calendar date.  So, in order to apply 'lost to censoring' you want to check censorWindow, not censorCriteria.  Confusing, I know.

PR #67, opened 2019-01-29
Added criteria groups to checking for unused concept sets

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1250

### DIFF
--- a/src/main/java/org/ohdsi/circe/check/checkers/UnusedConceptsCheck.java
+++ b/src/main/java/org/ohdsi/circe/check/checkers/UnusedConceptsCheck.java
@@ -61,7 +61,7 @@ public class UnusedConceptsCheck extends BaseCheck {
         }
         for (final ConceptSet conceptSet : expression.conceptSets) {
             boolean hasUsed;
-            if (!(hasUsed = isConceptSetUsed(conceptSet, Arrays.asList(expression.primaryCriteria.criteriaList)))) {                
+            if (!(hasUsed = isConceptSetUsed(conceptSet, Arrays.asList(expression.primaryCriteria.criteriaList)))) {
                 if (!(hasUsed = isConceptSetUsed(conceptSet, additionalCriteria))) {
                     for (InclusionRule rule : expression.inclusionRules) {
                         if (hasUsed = isConceptSetUsed(conceptSet, rule.expression)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/ohdsi/circe/check/checkers/UnusedConceptsCheck.java
```diff
@@ -51,14 +52,19 @@ protected WarningReporter getReporter(WarningSeverity severity, List<Warning> wa
     @Override
     public void check(CohortExpression expression, WarningReporter reporter) {
 
-        for(final ConceptSet conceptSet : expression.conceptSets) {
+        List<Criteria> additionalCriterion;
```
chrisknoll: Just a nitpick, but:  if it is a collection, then it's many things, so wouldn't it be _additionalCriteria_?  The singular would be criterion.  If it was in a for loop it would be `for (criterion in criteria) { }`

For simplicity, we always refer to these things as criteria becaues even in the 'single instance' case, the object is a collection multiple criteria (the age criterion, the gender criterion, etc).

I'd just like to avoid confusing the terminology.

src/main/java/org/ohdsi/circe/check/checkers/UnusedConceptsCheck.java
```diff
@@ -95,4 +101,14 @@ private boolean isConceptSetUsed(ConceptSet conceptSet, CriteriaGroup group) {
         return Objects.nonNull(criteriaList) ? Arrays.stream(criteriaList)
                 .map(c -> c.criteria).collect(Collectors.toList()) : Collections.emptyList();
     }
+
+    private List<Criteria> toCriteriaList(CriteriaGroup[] groups) {
+
+        List<Criteria> criterias = new ArrayList<>();
```
chrisknoll: Ah, yes here: 'criterias' isn't a word. The plural is 'criteria' the singular is 'criterion'.

aklochkova: Thank you. I didn't know about it

chrisknoll:
@aklochkova , there's one place you need to look for used concept sets: each criteria has the ability to have a CorrelatedCriteria associated with it: https://github.com/OHDSI/circe-be/blob/master/src/main/java/org/ohdsi/circe/cohortdefinition/Criteria.java#L52.  The CorrelatedCriteria field of Criteria is a CriteriaGroup, which I believe you have a check that will scan a criteria group and look for used concept sets.  So, I think the solution here is just to also check each Criteria's CorrelatedCriteria for unused concept sets.

chrisknoll:
In addition: I had to merge master into this PR so I could make sure I was testing with the latest commits to master. Please pull this branch again to get that merge.

PR #68, opened 2019-02-05
Adds warning when cohort inclusion criteria requires death before initial event

wivern:
Closes OHDSI/Atlas#1121

### DIFF
--- a/src/main/java/org/ohdsi/circe/check/Checker.java
+++ b/src/main/java/org/ohdsi/circe/check/Checker.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.List;
 import org.ohdsi.circe.check.checkers.ConceptSetCriteriaCheck;
 import org.ohdsi.circe.check.checkers.CriteriaContradictionsCheck;
+import org.ohdsi.circe.check.checkers.DeathTimeWindowCheck;
 import org.ohdsi.circe.check.checkers.DomainTypeCheck;
 import org.ohdsi.circe.check.checkers.DrugDomainCheck;
 import org.ohdsi.circe.check.checkers.DrugEraCheck;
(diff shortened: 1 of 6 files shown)


chrisknoll:
Hi, @wivern.  This logic is not correct.

You are using this field to determine your window:

```
		observationFilter = expression.primaryCriteria.observationWindow;
```

But, the primaryCriteria.observationWindow indicates how much continuous observation you must have before and after the cohort entry event:

image

However, what the issue is saying is that it should be a warning when you try to identify a Death event prior to the Cohort Entry Event, like this:

image

I understand the complexity here, but the basic rule is:  It is a warning if you attempt to identify Death prior to the start of an event (because the assumption is: if you die, you shouldn't expect to get any additional events.  The reason why this is a Warning vs. Error, it is possible that you might have post-death procedures (autopsy) that may identify conditions after death (tho this completely depends on how the ETL translated the data and what the data capture rules are). But since it's _possible_ to check for death prior to a diagnosis, _usually_ that's a logical error and the person should be warned.

pavgra:
@chrisknoll , could you try it once more?

chrisknoll:
Let's try to st the message from:

> inclusion rule Prior Death death criteria causes cohort cannot be created since a patient dies before being diagnosed with a condition

To: 

> inclusion rule Prior Death death criteria attempts to identify death event prior to index event.  Events post-death may not be available.

While the check did work looking for criteria in inclusion rules, it did not find the error when it was using the 'initial event restriction' (stored in 'Additional criteria'):
image

and it did not find it in nested criteria off the main cohort entry event:

image

It did, however, find the problem (correctly) when it was a nested criteria inside an inclusion rule:
image

So it is only in the nested criteria of primary criteria and within AdditionalCriteria that you also need to check this rule.

PR #70, opened 2019-02-08
Added condition to check attribute in Condition Occurrence

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1085


chrisknoll:
Hi, @aklochkova .  The other criteria types: Procedure, Drug, Measurement, and Observation also have source concept codesets, so I believe the check you applied to ConditionOccurrence should also apply to those other criteria types.

Issue #35, opened 2018-04-03
Indiscriminate use of 'invalid_reason is null'


schuemie:
All concept sets implicitly apply the requirement `invalid_reason is null`. However, this is not always desired. For example, I'm looking at source concepts for zoledronic acid injections using these concept IDs: 2718650, 2720787, 2718649, 44786564, 44786608. Only one of those appears to be currently valid, but the others are used a lot in the data, likely at a time in history when they were valid (or even if Christian says they weren't I would like to detect them). Currently, I have no option to include these concepts other than to manually modify the SQL.

Desired behavior: ability to define criteria using source concepts that were valid at the time, even if they no longer are valid now.

Actual behavior: source concepts that are currently no longer valid are automatically discarded.

Solution: Not sure. The whole reasoning for when a concept is valid or invalid is unclear to me. For source concepts the current behavior is definitely wrong, since the codes were valid in the past, and so at the very least the `invalid_reason is null` statement should be replaced with one testing whether the occurrence date is within the valid period. I would like to have the check for valid concepts disabled in this example altogether. I think these concerns might extend to standard concepts as well.

gowthamrao:
Lot of discussion about this here

http://forums.ohdsi.org/t/proposal-to-keep-outdated-standard-concepts-active-and-standard/3695

dimshitc:
and here's the final decision
http://forums.ohdsi.org/t/proposal-to-keep-outdated-standard-concepts-active-and-standard/3695/22

schuemie:
Thanks! In that case the `invalid_reason is null` statement would appropriately.

chrisknoll:
Re-opening this per request from https://github.com/OHDSI/ETL-CDMBuilder/issues/49.  Will remove invalid_reason checks from concept set queries.

PR #71, opened 2019-03-26
Invalid concept set are included in cohort generation(https://githubcom/OHDSI/WebAPI/issues/136)

ssuvorov-fls:
fixes https://github.com/OHDSI/WebAPI/issues/136

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
 	<groupId>org.ohdsi</groupId>
 	<artifactId>circe</artifactId>
-	<version>1.7.3-SNAPSHOT</version>
+	<version>1.7.4</version>
 
 	<properties>
 		<maven.compiler.source>1.8</maven.compiler.source>
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
 	<groupId>org.ohdsi</groupId>
 	<artifactId>circe</artifactId>
-	<version>1.7.4</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -6,7 +6,7 @@
 
 	<groupId>org.ohdsi</groupId>
 	<artifactId>circe</artifactId>
-	<version>1.7.3-SNAPSHOT</version>
+	<version>1.7.4</version>
```
wivern: Do You really want to make a release?

src/main/resources/resources/vocabulary/sql/conceptSetQuery.sql
```diff
@@ -1 +1 @@
-select concept_id from @vocabulary_database_schema.CONCEPT where concept_id in (@conceptIds)and invalid_reason is null
+select concept_id from @vocabulary_database_schema.CONCEPT where concept_id in (@conceptIds)
```
wivern: It's a bit concerning. I don't believe we want to select invalid concepts in all cases.
Probably it's better to make this optional.

chrisknoll: I have the same concerns here, but there was a thread on this at forums.ohdsi.org and they insist that the conceptSetQuery should return invalid concepts.  My fear is that we'll have a lot more results in the query, but this is what was decided (also, see #35 for details)

PR #75, opened 2019-04-10
Location region criteria

pavgra:
To support recently approved `location_region_id` column
https://github.com/OHDSI/CommonDataModel/issues/252


src/main/resources/resources/cohortdefinition/sql/locationRegion.sql
```diff
@@ -0,0 +1,21 @@
...
+  C.region_concept_id as TARGET_CONCEPT_ID,
+  CAST(NULL as bigint) as visit_occurrence_id
+from 
+(
+  select l.*, lh.*
+  FROM @cdm_database_schema.LOCATION l
+    JOIN @cdm_database_schema.LOCATION_HISTORY lh ON l.location_id = lh.location_id
+@codesetClause
+  WHERE lh.domain_id = (
+    SELECT CAST(concept_id AS VARCHAR)
```
chrisknoll: why do we need to cast a bigint field to varchar?

pavgra: @chrisknoll ,
image

chrisknoll: So, the concept_id in the query stores a value of either 'PERSON', 'PROVIDER' or 'CARE_SITE'? Those don't look like concept_ids.

chrisknoll:
Pavel, it is looking like you are rushing to a solution for CDM versioning handling. Please discuss these approach with the broader group before embedding this into a 'Location criteria' PR.

Ideally, we should solve the CDM version support problem, and then have this PR built on top of that function.

src/main/java/org/ohdsi/circe/cohortdefinition/LocationRegion.java
```diff
@@ -0,0 +1,19 @@
+package org.ohdsi.circe.cohortdefinition;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import org.ohdsi.analysis.versioning.CdmVersion;
+
+@JsonTypeName("LocationRegion")
+@CdmVersion(semver = ">=6.1")
+public class LocationRegion extends GeoCriteria {
```
chrisknoll: LocationRegionCriteria

chrisknoll:
Re: Location Criteria

I’ve tried to digest everything that was discussed via github and forums, the most detailed information was found here from Pavel:
https://github.com/OHDSI/WebAPI/issues/649

It feels like there’s still a bit of uncertainty about what the ‘full featured’ capability of geo data will be in the cdm (they just completed a vote on Location_distance), but Pavel, you presented a very nice roadmap, and I found the specific details about how you were thinking of implantation in circe:

Location criteria (circe-be)
The new Criteria to be introduced into circe:
- Person location is (not) in area
  -  Input:
      - area concept set
      - date range
      - should the person be inside the given areas or vice-versa?
- Person location is (not) within range from a care site
  - Input:
    - care site concept set
    - date range
    - GeoRange

image

Although, I’m not sure the PR you submitted actually implements this function.  It does contain the GeoCriteria (with startDate/EndDate) but none of the distance stuff, and instead has the notion of a ‘LocationRegion’ that just extends GeoCriteria to include a conceptSet (not described above).  I’m not familiar with the Vocabulary changes that would contain concepts that represent regional information, where is this described or does it exist?

While this PR covers getting regioinal information based on a Person, should it also support the location of a care_site and provider? Gowtham provided examples here:

**Requirement 3: Cohort definition using location table:**
1.	We would like circe-be to be enhanced to support location based criteria. Be able to select administrative areas as a criteria in circe-be + Atlas.
2.	Use administrative areas as part of cohort entry event or cohort inclusion criteria.
3.	Administrative areas must be derived from latitude and longitude in location table using look-up (point in polygon function, point is person's lat/long - polygon is the administrative area).
4.	Administrative area criteria may be specified thru text string search of the field city, state, zip, county, country of location table. This is similar to searching for LOT_NUMBER in drug_exposure as part of circe-be.
5.	We would like to calculate the distance of person.location_id from provider.location_id or care_site.location_id as part of cohort definition, e.g. We would like to build a cohort of subject_id who live more than 30 miles/kms from the provider.

Example in Atlas/circe-be: that should be supported at both cohort entry events (qualified events) and inclusion criteria (inclusion events) are
- where the care_site.location_id belonged to administrative area e.g. Lat/long was in the selected geographic area (in statement)
- where the person.location_id belonged to administrative area e.g. Lat/long was in the selected geographic area (in statement)
- where the care_site.location_id was within x units of distance from person.location_id (eucledian distance between two points defined by lat/long)

In all cases  above, the location_id is inferred by using the earliest valid location_id for the person_id/care_site_id/provider_id based on interval period of the qualified/inclusion event thru look-up to location_history

### Questions:

Using the current PR, which of the above cases can be handled by LocationRegion criteria? Is an ‘administrative area’ the same as a region conceptid?

Looks like the PR does support #1 and #2, if the purpose of the cohort entry is ‘person entered region’, or if you wanted to have a correlated criteria where person entered region between X and Y days of index event.

3-5 is about distances, I don’t think that’s possible (without location_distance).

The last 3 examples reference location_id for a care_site.  But the PR only involves the location info for the domain_id ‘person’.  Should we support the other domains?

pavgra:
@chrisknoll, the intention was to implement things step-by-step. So, this PR addresses the first part of the requirements, namely:

> Person location is (not) in area

Originally, I named the according criteria as `AreaGeoCriteria` but later, since the field name in CDM is `region_conept_id`, I named it `LocationRegion`. The appropriate CDM thread is https://github.com/OHDSI/CommonDataModel/issues/252.

The PR handles case of the location region as Inclusion rule (correlated criteria). Location region as cohort entry event was postponed since I've faced difficulties for combining observation periods and a location's start-end dates.

Care site's / provider's regions are not in the PR. Again, the intention was to implement it one by one.

Location distance is a separate criteria which I'm going to work out in a separate PR once the path for new criteria is clear.

src/main/java/org/ohdsi/circe/cohortdefinition/LocationRegion.java
pavgra: I see that 99% of criteria aren't postfixed with `Criteria`: `ConditionEra`, `DeviceExposure`, `Observation`...

PR #76, opened 2019-04-16
Implements Cohort Expression versioning

pavgra:
Depends on https://github.com/OHDSI/StandardizedAnalysisAPI/pull/22

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/PayerPlanPeriod.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/PayerPlanPeriod.java
@@ -19,8 +19,10 @@
 package org.ohdsi.circe.cohortdefinition;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.ohdsi.analysis.versioning.CdmVersion;
 import org.ohdsi.circe.vocabulary.Concept;
 
+@CdmVersion(range = ">=5.3")
 public class PayerPlanPeriod extends Criteria {
 	@JsonProperty("First")
 	public Boolean first;
--- a/src/test/java/org/ohdsi/circe/VersioningTest.java
+++ b/src/test/java/org/ohdsi/circe/VersioningTest.java
@@ -1,0 +1,76 @@
+package org.ohdsi.circe;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import org.junit.Test;
+import org.ohdsi.analysis.Utils;
+import org.ohdsi.circe.cohortdefinition.CohortExpression;
+import org.ohdsi.circe.cohortdefinition.CorelatedCriteria;
+import org.ohdsi.circe.cohortdefinition.Criteria;
+import org.ohdsi.circe.cohortdefinition.CriteriaGroup;
+import org.ohdsi.circe.cohortdefinition.PayerPlanPeriod;
+
+import java.io.IOException;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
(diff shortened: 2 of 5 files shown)


src/main/java/org/ohdsi/circe/cohortdefinition/PayerPlanPeriod.java
```diff
@@ -19,8 +19,10 @@
 package org.ohdsi.circe.cohortdefinition;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.ohdsi.analysis.versioning.CdmVersion;
 import org.ohdsi.circe.vocabulary.Concept;
 
+@CdmVersion(range = ">=5.3")
```
chrisknoll: Could we also apply a @cdmVersion(range = ">=5.0") on the base criteria class? Will that be overridden here?  I suggest this because I think it makes sense to declare that all criteria have some sort of `@CdmVersion` attached to them.

pavgra: You were typing it at the same time as I was commiting the change. Look below

src/test/java/org/ohdsi/circe/VersioningTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    public void checkSerializedVersionedPayerPlan() {
+
+        CohortExpression cohortExpression = new CohortExpression();
+        cohortExpression.additionalCriteria = getCriteriaGroupWithCriteriaList(getPayerPlanCriteria());
+        // User defined constraint
+        cohortExpression.setCdmVersionRange("<6.0.0");
+        String serialized = Utils.serialize(cohortExpression);
+        assertThat(serialized, containsString("\"cdmVersionRange\":\"^5.3.0\""));
```
chrisknoll: This test makes it appear that the user entered a value, but after serialization, the value is updated.  I'm not sure what the purpose of the test is.

Also: can we please add descriptions for each of these tests to specify what the test is trying to test?

pavgra: > I'm not sure what the purpose of the test is.

Hm, that the value is updated to the one which satisfies both user-defined and derived interval?..

chrisknoll: No, that isn't what we discussed.

During validation, the system scans all criteria instances to understand the overall ranges of CDMs that would be required. And with that range, you would verify that the user's intended CDM target was compatible with the ones based on derived range.  If not, we'd error out with the message of why.  And then the user has to double check that what they _intended_ to use as the target cdm actually is possible.  This was the part of our conversation:

Pavel: so should the circe throw an exception if it detects incompatibility between specified targetCDM and derived one?

I would go for throwing the exception as a safer option. And if we see that this extra safety somehow limits users, we can always remove it

1:40 PM
Yeah, throw an exception, and we can always implement Check classes that can throw up warnings about non-supported cdm version.

But, after all the conversation we had, I thought it was clear that we woudln't have Atlas 'think' for the user: if the user specifies a targetCDM range, and they do soemthing wrong, we shouldn't just 'fix' their targetCDM spec, we should alert them that their target spec covers an invalid range.

In the case of this spec, looks like the user has requested a targetVersion of <6.0.  However, there is stuff that became available after 5.3.0.  Therefore, the user shoudl be alerted that they have used features that are not available prior to 5.3.  The fix for the user is to say 'Ok, I can live with that, so i'll specify range of >= 5.3.0 < 6.0' or they can say 'I need it to work with everything < 6.0, so what's the problem? Oh payer plan period?  Ok, I'll remove it for this cohort definition, and make another one that is specific for >= 5.3'.

PR #80, opened 2019-04-19
Make CohortExpressionQueryBuilder more extensible

pavgra:
Current implementation of CohortExpressionQueryBuilder is difficult to extend due to the individual criteria sql builders are atomic functions that handle selects, joins and where clauses as atomic units.  In order to override any part of the criteria building, you need to copy over the remaining parts (provide a complete re-implemenation).

This PR proposes a structure that extracts the criteria building work to a separate, extensible class so that different elements of the criteria building can be overridden.

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
+import org.ohdsi.circe.cohortdefinition.builders.BaseCriteriaSqlBuilder;
 import org.ohdsi.circe.cohortdefinition.builders.LocationRegionSqlBuilder;
 import org.ohdsi.circe.cohortdefinition.builders.MeasurementSqlBuilder;
 import org.ohdsi.circe.cohortdefinition.builders.ObservationPeriodSqlBuilder;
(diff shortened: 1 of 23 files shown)


chrisknoll:
Thank you for your interest in proposing refactoring of the object model.  At this time, however, this is not an activity that has priority.  Feel free to continue development on this branch, and when bandwidth frees up to put the appropriate time and attention to this, we can revisit it.

In the meantime, I would either open an issue that describes the changes you intend to make, or add information to this PR with the same (the current information is 'No description provided'.

chrisknoll:
@pavgra, I updated the title and description to reflect the purpose of this PR.  Use it as a model for future PRs.

Initial comments:
I realize we discussed the encapsulation vs. inheritance, which led to the introduction of BuilderUtils to hold methods `getCodesetJoinExpression`, `getOperator()`, `dateStringToSql()`, `buildDateRangeClause()`, etc. instead of a base class that had those functions implemented only subclasses would access them.  At the time, the statements I made were more generalized about the accessibility of the method (only available to the subclasses vs. a 'utility' method that might me used outside of SQL building.

Now that I better understand the intent (which is to make CohortExpressionQueryBuilder more extensible), I'm leaning back towards having a 'BaseCriteriaSqlBuilder' class which is an abstract class, provides the default implementation of `getCriteriaSql()` 👍 
```
    public String getCriteriaSql(T criteria) {

        String query = VISIT_OCCURRENCE_TEMPLATE;

        query = embedCodesetClause(query, criteria);

        List<String> joinClauses = resolveJoinClauses(criteria);
        List<String> whereClauses = resolveWhereClauses(criteria);

        query = embedOrdinalExpression(query, criteria, whereClauses);

        query = embedJoinClauses(query, joinClauses);
        query = embedWhereClauses(query, whereClauses);

        return query;
    }
```
and the subclasses have the option to override the `embedXXX()` methods to provide a concrete implementation.

When deciding between 'should there be an interface or should there be an abstract class', I think about if the final class should have  multiple 'aspects' (by implementing multiple interfaces) or if the class extends an abstraction and 'fills in the blanks'.  I believe in the case of an abstract CriteriaSqlBuilder, the 'abstract base class' fits the pattern better.

So, my proposed changes are:
1. Put SqlBuilder suffix on the classes: VisitOccurrenceSqlBuilder and CriteriaSqlBuilder (the CriteriaSqlBuilder is the abstract class).
2. We could still keep a `BuilderUtils` class, but I think some of the methods that were put into BuilderUtils could belong to the abstract base class.  Basically, my judgement is 'If the method may need to be overridden by the child, we should put it into the base class'.  For example, `getOperator()` is something very specific to the behavior of CriteriaSqlBuilding and it is something that a child class may want to supply an extension of the `getOperator()` implementation itself...so this lends to subclassing.  `dateStringToSql` may be something that is a utility function, not really a defined 'behavior of the criteria builder that should be extended'...so in that case, I could see it being organized into `BuilderUtils`.
3. Since we have the abstract base class, no need to create a separate interface for 'CriteriaBuilder'. Just have an abstract base class 'CriteriaSqlBuilder'.  I think applying paramaterized types to the class will work the same way between an abstract base class vs. interface, so hopefully i'm not missing some other purpose of the paramaterized type.

I think that covers it.  I can see that the 'request-dispatch' pattern was not disabled in the PR: the CohortExpressionQueryBuilder continues to resolve the target method based on the subclass of Criteria by calling `accept()`, So I'm glad this was maintained.

src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java
```diff
@@ -1272,49 +1273,46 @@ public String getCriteriaSql(DrugExposure criteria)
...
-    return processCorrelatedCriteria(query, criteria);
+  public String getCriteriaSql(Measurement criteria) {
+    return getCriteriaSql(new MeasurementSqlBuilder<>(), criteria);
   }
   
   @Override
   public String getCriteriaSql(Observation criteria) {
-    String query = new ObservationSqlBuilder<>().getCriteriaSql(criteria);
-    return processCorrelatedCriteria(query, criteria);
+    return getCriteriaSql(new ObservationSqlBuilder<>(), criteria);
```
chrisknoll: I noticed that each of these getCriteriaSql create a new instance of the object for each invocation of the method.  Can they be allocated once and reused, or is it necessary that they get constructed for each invocation?

pavgra: @chrisknoll , addressed

PR #90, opened 2019-08-19
Issue 1933 atlas cohort check incorrectly flags concept set as unused

YaroslavTir:
Fixes https://github.com/OHDSI/Atlas/issues/1933

- concept set that used in the nested group was not checked and appears as worrying in the Cohort Message tab


src/test/java/org/ohdsi/circe/check/checkers/UnusedConceptsCheckTest.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.List;
+import org.junit.Test;
+import org.ohdsi.circe.check.Checker;
+import org.ohdsi.circe.check.Warning;
+import org.ohdsi.circe.cohortdefinition.CohortExpression;
+
+public class UnusedConceptsCheckTest {
+
+
+    private static final String EXPRESSION = "{\n" +
```
pavgra: I would move it into resources and read it from there

chrisknoll: Agreed, any content that would be loaded as a file, just make it an external resource (like our SQL templates are stored).

src/test/java/org/ohdsi/circe/check/checkers/UnusedConceptsCheckTest.java
```diff
@@ -0,0 +1,224 @@
...
+            "}";
+
+    private ObjectMapper objectMapper = new ObjectMapper();
+
+    private UnusedConceptsCheck unusedConceptsCheck = new UnusedConceptsCheck();
+
+
+    @Test
+    public void check_usageInTheChildGroups() throws IOException {
+        CohortExpression cohortExpression = objectMapper.readValue(EXPRESSION, CohortExpression.class);
```
chrisknoll: Use
```
  public static CohortExpression fromJson(String json) {
	return Utils.deserialize(json, new TypeReference<CohortExpression>() {});
  }
```

To deserlieize a string into a CohortExpression.

PR #92, opened 2019-08-28
Cohort generation fails on IMPALA due to wrong SQL #1312

YaroslavTir:
fixes https://github.com/OHDSI/WebAPI/issues/1312
wrap constant that are using in ORDER BY clause of analytic functions


chrisknoll:
Can you provide the example of the incorrect sql as well as the post-fix sql that this PR produces?

Thanks.

YaroslavTir:
Hi Chris,
sure this is an example of the incorrect sql(actually I keep only meeningful for this issue part):

```
select 
     row_number() OVER (PARTITION BY E.person_id ORDER BY e.start_date ASC) ordinal

from 
(
    select person_id,
           CAST('2011-01-01'  AS TIMESTAMP) as start_date
    from   synpuf_53_test.OBSERVATION_PERIOD 
) e


```
the problem here that we cannot put any constant in the `ORDER BY ` clause for IMPALA. this is intended behavior for this DB https://issues.apache.org/jira/browse/IMPALA-2775. This PR is like a quick fix, just for part that we know for sure having this issue. 

Of cause, this is not a proper way to fix this. The most general way is to make the SQL-Render finds this kind of constant and wrap them, or remove from them from order-by of analytic functions, but this really complicated solution, and can easily takes weeks of implementing. 

I also ponder about a more elegant solution for the circe. I want to not wrap such constants, but just remove them from ORDER_BY clause, but again, it also can be a quite complicated solution. So I would mark this PR as wip so far.

YaroslavTir:
@chrisknoll 
I have created an issue for a general fix, that I mentioned https://github.com/OHDSI/SqlRender/issues/187
Also, my idea to delete constants from `order by`, was not so good, this approach involves too much code changes.

anyway, after fix the original "buggy" code would look like this
```
select 
     row_number() OVER (PARTITION BY E.person_id ORDER BY e.start_date ASC) ordinal

from 
(
    select person_id,
        --add zero days to the constant
           DATEADD(dd, person_id-persion_id, CAST('2011-01-01'  AS TIMESTAMP) as start_date
    from   synpuf_53_test.OBSERVATION_PERIOD 
) e
```

chrisknoll:
I'm concerned about a performance impact of adding 0 days to a fixed date for each record just to get the fixed date into a non-constant state.

Also, I'm concerned about the readability of something like `person_id - person_id` as a hack to get impala to work with a 'non-constant constant'.

So, I think we can fix this on the circe-be side without bothering with SqlRender:  circe-be controls the row_number() call, and so we can dynamically produce the `ORDER BY` clause based on if s fixed-date was provided as input.  From your code exmaple,it look slike the only column it orders by is the date, but we can make this order by the observation_period_id column....

In the case where there isn't a user-defined date, we can order by observation_period_start_date, observation_period_id...in the case where there is a user-defined date, we can just order by observation_period_id.  Effectively removing 'constants' from the order by, without having to make sqlRender figure it out.

YaroslavTir:
@chrisknoll   hmm, you know I like your idea. I guess if it possible to fine good alternative of the "constant" fields, and change query accordingly, then it would be the best solution. Can we really trust observation_period_id column as an alternative for start_date?  I know that hibernate generate id just increment the last id value. Is it possible that we put value to the table in the "wrong" order?

chrisknoll:
Good question, I was thinking the same thing, but what I was currently ordering on was the user-defined date, which would be a constant, so none of the records would have any logical order anyways.

So, another suggestion, and I think this one will work:

Let's pull 2 dates from the subquery: a `start_date` and a `sort_date`. The `sort date` will always be the observation_period_start_date.   The `start_date` will either be the observation_period_start_date, or it will be the user defined date.   In our `row_number()` statement, we always will `partition by persion_id, order by sort_date`.  This way, we never will sort on a constant.

If this is confusing, I can submit a different PR to demonstrate this. Let me know if it's not clear.

YaroslavTir:
@chrisknoll Indeed I added `sort_date` field for all criteria queries, and put it in the `order by` clause  I checked a few cases, they work like a charm. Thank you for a good suggestion  :) tbh I was thinking about something like this, but was afraid that it would entail too many changes, in fact, it turned out not so bad.

src/main/resources/resources/cohortdefinition/sql/locationRegion.sql
```diff
@@ -5,7 +5,8 @@ select
   C.start_date as start_date,
   ISNULL(C.end_date, DATEFROMPARTS(2099,12,31)) as end_date,
   C.region_concept_id as TARGET_CONCEPT_ID,
-  CAST(NULL as bigint) as visit_occurrence_id
+  CAST(NULL as bigint) as visit_occurrence_id,
+  C.start_date as sort_date,
```
pavgra: Doesn't seem that comma is needed here. Doesn't it fail?

YaroslavTir: fixed

chrisknoll:
Thank you for the comprehensive fix.  I was thinking only payer plan period and observation period criteria would need the fix (since that's the place where you can specify a constant for the start date), but this way makes it more uniform.

Let me run a few tests and if it passes, I can merge the updates.

PR #93, opened 2019-09-01
Makes cohort id column name configurable

pavgra:


chrisknoll:
I added tests and made the default cohort id field population consistent (not depending on sqlRender to set it inn one case, and using options in the other case).

@pavgra , do not force push over these changes.

@anthonysena : these changes may also impact your negative control functionality, please provide your review on this PR as well.

pavgra:
@chrisknoll , what do you think of applying some consistent formatting?

This looks a bit sad
image

chrisknoll:
@pavgra , as we've discussed in the past,  we've made the decision in WebAPI to adopt tab indentation, so these changes were applying that formatting, but only touching the lines changed for a given PR (so not to have too much whitespace noise).

However, this repo has, from the beginning, used 2 space indent.  So, we'll adopt that formatting here, so it is consistent within this repo.  I'll adjust the files touched in this PR to ensure the 2 space indentation.

anthonysena:
Thanks @chrisknoll for making these changes. I'll do some tests of the negative control cohorts and leave any feedback here.

anthonysena:
@chrisknoll no concerns about these changes from my side. If we ever need to move in this direction for negative control cohorts I can augment my code to take advantage of the parameterized cohort_id column.

PR #96, opened 2019-09-30
Fixes cohort expressions with nested criteria

pavgra:
fixes https://github.com/OHDSI/circe-be/issues/95

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
 	<groupId>org.ohdsi</groupId>
 	<artifactId>circe</artifactId>
-	<version>1.8.2-SNAPSHOT</version>
+	<version>1.8.3-SNAPSHOT</version>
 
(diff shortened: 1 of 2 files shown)


chrisknoll:
Don't merge PRs into released branch. That's just a commit ref pointer for jenkins build.

Since you've already set up a rc branch for 1.8.3, we can just use that.

pavgra:
I thought that: Jenkins builds off the `released` branch so that code should be there + we want the 1.8.3 release with the fix = PR into the released

chrisknoll:
Released branch gets updated once we make a release that we want to publish as a 'release'.  Once the 1.8.3 version of the library is released via git (and a v1.8.3 tag is applied) the release branch is moved to the v1.8.3 tag.  We don't create new releases every time a new commit comes into the repo.

PR #102, opened 2019-11-26
Fixed warnings.

chrisknoll:
Fixed warnings.
Fixed NC tests.


chrisknoll:
@anthonysena ,
 As i was prepping my branch for adding DB test cases, I found a lot of warnings I wanted to resolve, and ensure that all tests continued to function.  

I found that the NC tests were breaking due to some changes related to the new `cohort_id_field-name` options.  Could you review these changes and let me know if there's any problem?  The main focus is that I only put the `cohort_id_field_name` at the location where the insert to the final cohort table was being applied.  Everything used 'below' taht point, doesn't need to have a custom field name applied to it (the lower-level queries can refer to thigns as cohort DI without breaking the custom cohort_id_field_name.  But, there may be a case where a domain query is somehow used directly outside of the NC cohort INSERT statements, so...just please check my logic here and let me know if there's any problems.

anthonysena:
@chrisknoll - thanks for noting this. I reviewed the negative control queries and this all looks good from my side. You are correct - the sub-queries did not need to use the `cohort-id-field-name` mask only the top-level insert.

PR #105, opened 2019-12-18
Fixes SQLRender default param usage

wivern:
Fixes OHDSI/circe-be#104


chrisknoll:
@wivern , this should be unnecessary because those parameter replacements are handed via the options passed into build, shown [here](https://github.com/OHDSI/circe-be/blob/master/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java#L362).

I haven't had time to address it, but I plan on removing those 'defaults' from the sql.

wivern:
Well, so I guess we shouldn't concern about that if `defaults` would be removed.
Just I'd like to note that until defaults have gone the hydrated package of estimation analysis produces invalid sql for negative control outcomes. 
```
org.postgresql.util.PSQLException: ERROR: syntax error at or near "{"
  Position: 1

SQL:
{default @cohort_id_field_name cohort_definition_id}

INSERT INTO temp.temp_cohort_ixiie854 (
	subject_id,
	@cohort_id_field_name,
	cohort_start_date,
	cohort_end_date
)
SELECT
	s.subject_id,
	s.cohort_definition_id,
	s.cohort_start_date,
	s.cohort_start_date cohort_end_date
```

chrisknoll:
Yes, negative control SQL may work slightly different than how the cohort expression builder works.

So, on that note, I think that this is fine from a negative controls perspective.  I'll speak for @anthonysena that I think these changes are good (in that they do fix the `default` syntax issue.

PR #107, opened 2020-01-16
checks of the presence of values for the parameters were added

ssuvorov-fls:
fixes https://github.com/OHDSI/Atlas/issues/387
now concept array parameters are checked
demographic criteria are also now checked


chrisknoll:
I checked the coverage report here: https://codecov.io/gh/OHDSI/circe-be/pull/107/diff, and it looks like there are some minor gaps in the test coverage of some newly tracked files. Could you review that report, and adjust tests so that your PR is completely covered?

ssuvorov-fls:
> I checked the coverage report here: https://codecov.io/gh/OHDSI/circe-be/pull/107/diff, and it looks like there are some minor gaps in the test coverage of some newly tracked files. Could you review that report, and adjust tests so that your PR is completely covered?

Sure

chrisknoll:
Thanks, @ssuvorov-fls , that last commit increased the coverage, but there's one more type of test you should add that should cover a few of the gaps in BaseValueCheck, which is to test when you pass in a NULL value as the criteria. Can you add a null test to ensure all  the null checks are working as expected.

That should cover it, and then the only outstanding comment is made by @wivern. He's requesting an empty criteria check (as warnings).  Do you think that falls into the scope of this PR or should we have a separate issue raised for 'empty criteria checks'?

ssuvorov-fls:
@chrisknoll 
Hello, I've made some refactoring and ide shows me 100% coverage, but  https://codecov.io/gh/OHDSI/circe-be/pull/107/diff still shows less than 100%.

chrisknoll:
@ssuvorov-fls 
The remaining gaps are in the branches like here:
https://codecov.io/gh/OHDSI/circe-be/pull/107/src/src/main/java/org/ohdsi/circe/check/checkers/BaseValueCheck.java#D1-72

```
if (Objects.nonNull(expression)) {
  for (final InclusionRule rule : expression.inclusionRules) {
    if (Objects.nonNull(rule.expression)) {
```

The first and third line above are showing partial coverage because there was no test case where the expression or rule.expression were null.  All the test cases are built with non-nulll expressions, so that branch of logic isn't covered.

This is a minor thing, but I was going to try to add a commt that submits a null criteria to the checks to see if it will cover those lines.  That's what I meant in my above response about 'null checks'.

I was planning on doing this on 1/21 (after our holiday here) so I can try it and see if it increases the coverage.  Otherwise, 94% is decent coverage and we can go with that, but I do want to see if we can close the gap on those null cases.

chrisknoll:
Ok, with 97% coverage, I'll accept it. Thanks very much for the additional coverage.

chrisknoll:
The only remaining open item is @wivern's comment about empty criteria checks.  Is this something that should be done here, or is it a separate issue?

@wivern  and @ssuvorov-fls , please coordinate the remaining action item.

wivern:
@chrisknoll Sergei has made changes I requested, so nothing prevents to merge this from my side

PR #110, opened 2020-02-19
Issue 2129 prevent sql injection fix

YaroslavTir:
Fixes https://github.com/OHDSI/Atlas/issues/2129


YaroslavTir:
Fixes https://github.com/OHDSI/Atlas/issues/2129
Small investigation shows that validation/escaping of the params is not the best approach in the first place, but it is applicable if we provide SQL to the third party. I have read https://owasp.org/www-project-cheat-sheets/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet documentation and check the code. This framework escaping SQL params and do it differently for different dialects. I have found the common part for all dialects and this part really straightforward: just double single quote and escaped single quote

src/main/java/org/ohdsi/circe/cohortdefinition/builders/BuilderUtils.java
```diff
@@ -136,9 +136,17 @@ public static String buildTextFilterClause(String sqlExpression, TextFilter filt
...
+        String value = escapeSqlParam(filter.text);
+
         return String.format("%s %s like '%s%s%s'", sqlExpression, negation, prefix, value, postfix);
     }
+
+    private static String escapeSqlParam(String value) {
+        if (StringUtils.isEmpty(value)) {
+            return value;
+        }
+        return value.replaceAll("'", "''");
```
chrisknoll: In your unit test, you said that 'Simply doubling quotes doesn’t work', but looking at this code, it looks like you are just replacing single quotes with double quotes.   And, i don't see where this code applies the same logic as your test.

Can you confirm that you're fixing the sql injection?

chrisknoll:
The build is failing on this branch:
```
[ERROR] /home/travis/build/OHDSI/circe-be/src/test/java/org/ohdsi/circe/cohortdefinition/builders/BuilderUtilsTest.java:[7,31] package org.apache.commons.lang does not exist
[ERROR] /home/travis/build/OHDSI/circe-be/src/test/java/org/ohdsi/circe/cohortdefinition/builders/BuilderUtilsTest.java:[17,51] cannot find symbol
  symbol:   variable StringUtils
  location: class org.ohdsi.circe.cohortdefinition.builders.BuilderUtilsTest
[ERROR] /home/travis/build/OHDSI/circe-be/src/test/java/org/ohdsi/circe/cohortdefinition/builders/BuilderUtilsTest.java:[32,52] cannot find symbol
  symbol:   variable StringUtils
  location: class org.ohdsi.circe.cohortdefinition.builders.BuilderUtilsTest
```

I believe we already are using some dependency on org.apache.commons.lang, so please do not introduce a new dependency if we already have string manipulation classes to support a function.

src/main/java/org/ohdsi/circe/cohortdefinition/builders/BuilderUtils.java
YaroslavTir: @chrisknoll you are absolutely right! This is probably a bad excuse, but I did not commit the latest changes. anyway, this case is covered and unit test is green.

PR #113, opened 2020-03-24
Reduces number of inserts

wivern:
Fixes OHDSI/SqlRender#208


chrisknoll:
@wivern, let's do it this way:

Create the fix on the master branch where I have test cases built and the CI env will test the change.

If tests pass,  we can cherry pick the master commit onto the 1.8.5-hotfix branch.

In addition, I'm going to put a commit on the 1.8.5-hotfix branch to incremnt the pom.xml version so you don't have to do it in your commit.

PR #116, opened 2020-03-26
fix: the wrong insert clause is generated in case empty concepts list.

YaroslavTir:
the insert was optimized by replacing multiply ` insert-into-selects` with one `insert` and all `select` was grouped by the `union`, but in the case of an empty list the insert clause was invalid:
`insert (a,b);`.


chrisknoll:
Hi, am I correct that the current set of test cases didn't cover this use-case (I'm assuming such since the original PR did not flag an error). I should have asked that the test case be written to cover it.

Please add a test case in the CDMV5 that exercises an empty cohort set scenerio.

YaroslavTir:
> Hi, am I correct that the current set of test cases didn't cover this use-case (I'm assuming such since the original PR did not flag an error). I should have asked that the test case be written to cover it.
> 
> Please add a test case in the CDMV5 that exercises an empty cohort set scenerio.

this a really good remark. I added unit tests. I also replaced foreach with the stream, I guess this a more readable way

PR #123, opened 2020-06-29
Added hashcode and equals methods

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/1538

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/ConceptSet.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/ConceptSet.java
@@ -18,6 +18,7 @@
  */
 package org.ohdsi.circe.cohortdefinition;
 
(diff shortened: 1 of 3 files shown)


chrisknoll:
Hello, here's how I'd like to proceed with PRs in CIRCE:

All main bug fixes get merged into master.  If we need a hotfix (for a 1.8.x for example) we can cherry pick commits out of master into a hotfix branch and deploy those hotfixes as we deem necessary.

What I'm trying to avoid is missing hotfixes, and it feels it's safer to merge fixes into master (where we actually have test cases) and then cherry pick off master into a hotfix branch vs. pusing stuff from a hotfix branch into master (and sometimes pushing into hotfix branch from master).  I feel it will be easier to manage if we go in one direction PR->master->cherry pick to hotfix.

PR #124, opened 2020-06-30
Added hashcode and equals methods

aklochkova:


chrisknoll:
Let's implement a few test cases to cover those new methods.  I suggest the easiest approach is to add a .json serialization of a concept set into https://github.com/OHDSI/circe-be/tree/master/src/test/resources under a new folder 'conceptsets', and the test case is to read the same file twice and perform the `.equals()` operation on it. Then you can load a separate .json expression that is different, and confirm that the equals does not match.

chrisknoll:
I added a commit to address the following items;
- ConceptSet's code coverage had some gaps, so I added test cases where we check for nulls and different objects.
- Indentation was not aligned with other files in circe: this repository uses 2-space indent.

PR #127, opened 2020-09-03
"Fix it" for IncompleteRuleCheck does not work in Atlas #126

### LINKED ISSUE #126: "Fix it" for IncompleteRuleCheck does not work in Atlas
"Fix it" button for incomplete rule check should remove the rule by it's name.
изображение

Cause of the problem: "ruleName" field from backend does not contain the rule name but this:
изображение

anton-abushkevich:
Fixes #126

### DIFF
--- a/src/main/java/org/ohdsi/circe/check/checkers/IncompleteRuleCheck.java
+++ b/src/main/java/org/ohdsi/circe/check/checkers/IncompleteRuleCheck.java
@@ -44,7 +44,7 @@ public class IncompleteRuleCheck extends BaseCheck {
 
 	private void checkInclusionRule(InclusionRule rule, WarningReporter reporter) {
 		if (rule.expression.isEmpty()) {
-			reporter.add(rule.name + " has no criteria");
+			reporter.add(rule.name);
 		}
 	}
 }


chrisknoll:
Could this work off of a 'rule index' instead of a 'rule name'? I'm uncomfortable by keying elements by a user-input field, although I think this will work, it just doesn't feel clean.  We also don't have any restrictions on giving different rules the same name....

Is there a limitation in the reporter framework where it requires just a string identifier to locate the element?

I'm thinking about enhancing the object model to add a field like 'refId' so that you can target specific elements for warnings....my main holdup on this is that I don't really want to pollute the json object model with refIds when it has nothing to do with the actual expression and it may confuse the readers (of the JSON) as to what those reference IDs mean (when they are going to be arbitrary-but-unique identifers for each node in the object model).  The other idea was to associate a 'path' element to the reporter so that if you find a problem with a criteria it would be somethign like 'primaryEvents[0]' means the first entry criteria, while 'InclusionRules[1]' means the second inclusion rule, and 'InclusionRules[1]/criteriaList[3]' means the 4th criteria element of the second Inclusion rule...you can see where I'm going with this, and it can be a little complicated.  

Using unique reference IDs, we don't have to work out how to navigate to an element via a path, you just seek it in the object model for the target refId...but like I said earlier, it pollutes the object model....so, i'm just wrestling with this, but it would be good to have a formal function that allows targeting specific expression elements....

Open to suggestions!  Feel free to share your thoughts.

anton-abushkevich:
@chrisknoll, I agree that using user input fields as keys is not so great idea, but if choose between it and more complex model, I better choose former. I think clean model has priotity in this case. Better to add a restriction on using the same name for different Inclusion Rules than to complicate the model with additional identifiers/indices. 

The path idea looks interesting, but does it worth it? Maybe postpone it until next major version of Atlas? What we have now is more or less covers verification needs.

chrisknoll:
So the other way to handle this without polluting the model is to introduce a mechanism of traversing the expression tree (which current happens when we traverse the elements looking for errors), we annotate the warning to the sequence ID of the element that we traversed.  Then, on the front end if we need to locate the element that the warning refers to, we traverse the same expression object graph on the client (javascript) side to locate the referenced element...so all that has to be in sync is the server-side and client-side traversal to find the right element the warning refers to.  Of course, one headache is that if the object changes structure on the client side, the the warning messages from the server side are going to be out of sync (but we can manage this by 'invalidating' the warning messages if the expression 'is dirty' compared to the last time the warning messages were processed.  Generating a uniqe element ID in the model wouldn't have this problem, but attaching a path to the warning would:  if you remove a criteria or a group from the expression, the path (like criteriaList[1] is now criteriaList[0]) is invalid now....

Ok, so this is a lot of information, and let me express why I'm giving this so much attention:  my worry is that if we implement a solution that depends on 'unique names' for elements, we can only report issues (and indicate the element in the UI) for only those things that have names (in this case, concept sets and inclusion rules).  But longer term, if we find a criteria element that has a validation problem, it would be good to be able to either highlight or show in a popup (from the messages tab) the snipit of the criteria object that has the problem.  Doing this by name puts us into a corner.  

So, this is somethign I had planed on addressing in a 1.9 release (since it woudl be a new feature) but  what I feel like we're doing here is more of a hotfix of 1.8. ... but we're going to the 1.9 version for atlas 2.8 so...I'm having a hard time doing this quck-and-dirty when we have an opportunity to handle long term problems without going further down the 'wrong path' of depending on element names....

That's my thoguhts. Again, I don't want to be a bottleneck for progress, but I do feel like we need a more well-formed solution long term.

chrisknoll:
@anton-abushkevich , for the porposes of not letting this linger, let's go with your approach for this fix, but I ask that you add a test case for this PR...just a simple create an invalid JSON with this warning, and then check to see that you get the correct warning.  Could you do that?  Then we can close this out and move on, and i'll propose a more formal approach to identify elements in the object model so that we can have targeted warnings for our cohort expression.

chrisknoll (approved):

PR #130, opened 2020-09-16
critical error message when using all events per person is now warning

### LINKED ISSUE #129: Invalid CRITICAL error message when using all events per person
When selecting 'all events' for initial events, but then 'earliest event' for 'initial event restrictions' or 'qualifying events', a critical message is raised that you must specify an exit strategy when all events per person is specified. This prevents generation of a perfectly fine cohort.

### Steps to reproduce:
1. Import attached JSON file
2. Observe messsages saying : "all events" are selected and cohort exit criteria has not been specified, but earliest event per person is specified twice!

### Additional notes

I think this should be a warning, even if someone specifies all events and does not specify exit criteria.  Yes, it will be inefficient, but it will generate and you will get a (correct) result.  We should limit CRITICAL messages to those things which we know will fail execution.  Anything else is a warning or info.

[invalidWarning.txt](https://github.com/OHDSI/circe-be/files/5226716/invalidWarning.txt)

ssuvorov-fls:
fixes #129

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
           <!-- Sets the VM argument line used when unit tests are run. -->
           <argLine>${surefireArgLine}</argLine>
           <!-- Skips unit tests if the value of skip.unit.tests property is true -->
-          <skipTests>${skip.unit.tests}</skipTests>
+          <skipTests>true</skipTests>
           <!-- Excludes integration tests when unit tests are run. -->
           <excludes>
               <exclude>**/IT*.java</exclude>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
           <!-- Sets the VM argument line used when unit tests are run. -->
           <argLine>${surefireArgLine}</argLine>
           <!-- Skips unit tests if the value of skip.unit.tests property is true -->
-          <skipTests>true</skipTests>
+          <skipTests>${skip.unit.tests}</skipTests>
           <!-- Excludes integration tests when unit tests are run. -->
           <excludes>
               <exclude>**/IT*.java</exclude>
--- a/src/main/java/org/ohdsi/circe/check/checkers/NoExitCriteriaCheck.java
+++ b/src/main/java/org/ohdsi/circe/check/checkers/NoExitCriteriaCheck.java
@@ -21,11 +21,19 @@ package org.ohdsi.circe.check.checkers;
 import static org.ohdsi.circe.check.operations.Operations.match;
 
 import java.util.Objects;
+
+import org.ohdsi.circe.check.WarningSeverity;
 import org.ohdsi.circe.cohortdefinition.CohortExpression;
 
 public class NoExitCriteriaCheck extends BaseCheck {
 
     private static final String NO_EXIT_CRITERIA_WARNING = " \"all events\" are selected and cohort exit criteria has not been specified";
+
+    @Override
+    protected WarningSeverity defineSeverity() {
+
+        return WarningSeverity.WARNING;
+    }
 
     @Override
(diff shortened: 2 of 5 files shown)


chrisknoll:
I think you made the warning change (so that the cohort can generate) but the core of the problem is the warning is wrong: there should not be a warning that 'all events' are set because there are other places that say 'earliest event per person' so we aren't getting > 1 event per person.

The three places we should check are:
CohortExpression.primaryCriteria.primaryLimit
CohortExpression.qualifiedLimit **note**
CohortExpression.expressionLimit

**note**: qualfiedLimit is only applied when there is a non-null in CohortExpression.additionalCriteria, so, the logic is:
```
if (expression.primaryCriteria.primaryLimit.type.equalsIgnoreCase("ALL") && 
  (expression.additionalCriteria == null || !expression.qualifiedLimit.type.equalsIgnoreCase("ALL")) &&
  !expression.expressionLimit.type.equalsIgnoreCase("ALL") &&
  expression.endStrategy == null) {
   // warning:  no cohort exit defined when using all events per person.
}
```

chrisknoll:
In addition, could you apply a test case for this?

pom.xml
```diff
@@ -64,7 +64,7 @@
           <!-- Sets the VM argument line used when unit tests are run. -->
           <argLine>${surefireArgLine}</argLine>
           <!-- Skips unit tests if the value of skip.unit.tests property is true -->
-          <skipTests>${skip.unit.tests}</skipTests>
+          <skipTests>true</skipTests>
```
chrisknoll: how come this was hard-coded to a value instead of using the property?

src/main/java/org/ohdsi/circe/check/checkers/NoExitCriteriaCheck.java
```diff
@@ -37,9 +37,12 @@ protected WarningSeverity defineSeverity() {
...
     @Override
     protected void check(CohortExpression expression, WarningReporter reporter) {
-
         match(expression)
-                .when(e -> Objects.equals("All", e.primaryCriteria.primaryLimit.type) && Objects.isNull(e.endStrategy))
+                .when(e -> Objects.equals("All", e.primaryCriteria.primaryLimit.type)
+                        && Objects.isNull(e.endStrategy)
+                        && Objects.equals("All", e.expressionLimit.type)
+                        && (Objects.isNull(e.additionalCriteria)
+                            || Objects.equals("All", e.qualifiedLimit.type)))
```
chrisknoll, line 45: we use 'equalsIgnoreCase' in many places, i apologize that we didn't put these coded values in some kind of constants, but that's the world we're currently living in. I plan on remediating some of this in Circe 2.0.

pom.xml
ssuvorov-fls: i fixed it in second commit

chrisknoll:
I think one of my comments was out of date, apologies.  

Last thing I'd ask is to check the indentation on these files: circe has adopted 2-space indents from the start (unlike WebAPI and possibly other repos you've worked with).  I do not want to mix 2-space and 4-space indentation or even tab-based (you're not using any tab indents, as far as I can tell, so that's good).  It looks like your new test classes is using 4-space indents, please make that 2-space.  The test .json file you added is 2-space indent, so that looks good.

src/main/java/org/ohdsi/circe/check/checkers/NoExitCriteriaCheck.java
ssuvorov-fls: fixed

ssuvorov-fls:
indentation was fixed for check and test classes

chrisknoll (approved):

PR #177, opened 2022-09-23
Remove all CTEs from generateCohortSql.

### LINKED ISSUE #172: performance of CTEs on mpp platforms
In working with AWS technical support they have identified an issue with how some CTEs are used in the generated SQL.  We have identified some potential approaches to improve performance and will create a branch for review once all their findings are available. In one particular finding the use of a CTE is preventing queries from completing in cases where there are very large (billions) record counts in tables that are used in multiple joins.

chrisknoll:
Fixes #172.


chrisknoll:
@AnthonyMolinaro : if you want to verify this change, I can take in a .json expression and generate the new SQL for you to verify. Alternatively, I can create a branch of circe-r that has the updated code that you can generate sql yourself.

AnthonyMolinaro:
@chrisknoll  Great work, Chris!!

chrisknoll:
Just to close the loop: I took a sample of 30 cohorts and compared execution times:

image


Overall, the red (no cte) is faster than blue (original code), but there are a few case where red is slower than blue. However, when it is slower, it's not much slower, and there are much more cases where the non-cte approach is much faster.

In addition, this isn't a great scientific study because the queries ran at different times of the day with different background activity.  However, I think in general it's an improvement.

PR #185, opened 2023-01-13
Forces decimal formatting to Locale.US - fixes #184

### LINKED ISSUE #184: Decimal formatting based on Java locale
Currently when circe-be translates from a JSON specification to SQL, it uses the default locale of the machine to perform the translation. This is problematic in the case when the locale is German since this locale formats a decimal place with a "," instead a ".".

Under these circumstances, a SQL error may be generated. For example, if we were to look for a measurement value < 10, the SQL would be formatted as:

```sql
from 
(
  select m.* 
  FROM omopcdm.MEASUREMENT m
JOIN Codesets cs on (m.measurement_concept_id = cs.concept_id and cs.codeset_id = 16)
) C
WHERE C.value_as_number < 10,0000
```

instead of

```sql
from 
(
  select m.* 
  FROM omopcdm.MEASUREMENT m
JOIN Codesets cs on (m.measurement_concept_id = cs.concept_id and cs.codeset_id = 16)
) C
WHERE C.value_as_number < 10.0000
```

This is a relevant discussion on StackOverflow related to decimal formatting: https://stackoverflow.com/questions/5054132/how-to-change-the-decimal-separator-of-decimalformat-from-comma-to-dot-point.

I'm not sure how we'd want to handle this: either trying to force the locale via Java (which may be hard to do in R) or if we should just assert this in circe-be directly since the target of the translation is ANSI-SQL?

anthonysena:
Provides handling for formatting doubles to force formatting to the US locale.

### DIFF
--- a/src/main/java/org/ohdsi/circe/cohortdefinition/builders/BuilderUtils.java
+++ b/src/main/java/org/ohdsi/circe/cohortdefinition/builders/BuilderUtils.java
@@ -6,6 +6,7 @@ import org.ohdsi.circe.cohortdefinition.NumericRange;
 import org.ohdsi.circe.cohortdefinition.TextFilter;
 import org.ohdsi.circe.vocabulary.Concept;
 
+import java.util.Locale;
 import java.util.ArrayList;
 
 public abstract class BuilderUtils {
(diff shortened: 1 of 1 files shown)


anthonysena:
Let me get confirmation that this fix works from those collaborators that encountered the issue. Then I'll mark this as ready for review. Thanks!

chrisknoll (approved):

PR #190, opened 2023-05-02
Upgraded maven libs

ecruz165:
Upgraded maven libs and added spring boot parent to manage spring lib dependency versions.


chrisknoll:
So, this library (circe-be) is not a spring application at all, and tho there may be some shared dependencies that you might find in a spring boot application that you see here (such as JSON serialization), this definitely is not something that is placed into a Spring container (where you define `@Components` `@service` etc.  So, I'm onlly have the cocnern about adding spring boot to this POM insofar that we're no ta spring application.

Does that make sense?

chrisknoll (approved):

PR #200, opened 2023-12-21
Support for datetime calculations in cohort definitions

jcnamendiOdysseus:


src/main/java/org/ohdsi/circe/check/checkers/Comparisons.java
```diff
@@ -31,6 +29,12 @@
 
 public class Comparisons {
 
+    private static final Map<String, Integer> TIME_UNIT_CONVERSION = new HashMap<>();
+    static {
+        TIME_UNIT_CONVERSION.put(IntervalUnit.HOUR.getName(), 60 * 60);
+        TIME_UNIT_CONVERSION.put(IntervalUnit.MINUTE.getName(), 60);
+    }
+
```
chrisknoll, line 37: I wouldn't do it this way, it makes it seem like it's dynamic (ie: can change at runtime). Can't this function of conversion be associated to the enum itself?

alex-odysseus:
Chris, please don't go deep into this pull request as it is a draft and incomplete @chrisknoll

chrisknoll:
Will be easier to point out issues earlier instead of sifting through potentially thousands of lines of changes.

.editorconfig
```diff
@@ -8,4 +8,4 @@ root = true
 [*.{java,sql}]
 charset = utf-8
 indent_style = space
-indent_size = 2
+indent_size = 4
```
chrisknoll, line 11: don't change indents, 2 space indent is the norm. While there are some files that were introduced by ODY that have 4sp indent, the standard is 2 space for CIRCE.

src/main/java/org/ohdsi/circe/check/checkers/Comparisons.java
```diff
@@ -101,19 +106,26 @@ public static Predicate<Concept> compare(Concept source) {
...
-            range2End = window.end.coeff * window.end.days;
+        int range1 , range2Start , range2End;
+        if(Objects.nonNull(window.start) && Objects.nonNull(window.start.days)){
+                range1 = filter.postDays + filter.priorDays;
+                range2Start = window.start.coeff * window.start.days;
+                range2End = Objects.nonNull(window.end) && Objects.nonNull(window.end.days) ? window.end.coeff * window.end.days : 0;
+        }else {
+                range1 = (filter.postDays + filter.priorDays) * 24 * 60 * 60;
+                range2Start = getTimeInSeconds(window.start);
+                range2End = getTimeInSeconds(window.end);
```
chrisknoll, line 118: This is where the 4sp indent is glaring, so before you have to go back and change all of it back to 2sp, i'd adjust your formatting.

src/main/resources/resources/cohortdefinition/printfriendly/conceptSet.ftl
```diff
@@ -1,22 +1,22 @@
...
-|Concept ID|Concept Name|Code|Vocabulary|Excluded|Descendants|Mapped
-|:---|:---------------------------------------|:--|:-----|:--:|:--:|:--:|
-<#items as conceptSetItem>
-|${conceptSetItem.concept.conceptId?c}|<#--
+    ### ${conceptSet.name}
+    <#list conceptSet.expression.items>
+        |Concept ID|Concept Name|Code|Vocabulary|Excluded|Descendants|Mapped
+        |:---|:---------------------------------------|:--|:-----|:--:|:--:|:--:|
+        <#items as conceptSetItem>
+            |${conceptSetItem.concept.conceptId?c}|<#--
```
chrisknoll, line 8: You want to be extremly careful with changing the formatting of the Freemarker templates (ftl files) that are building the markdown format, since these are very sensitive to whitespacing.  Unless you need to change this, I would not touch any of these.

.editorconfig
jcnamendiOdysseus: Fixed

TomWhite-MedStar:
@chrisknoll , if this pull request is related to https://github.com/OHDSI/Atlas/issues/2886, I'm looking for advice on getting past a hurdle.  the GUI looks good, and when one explicitly specifies hour or minute for a widget, the generated code looks correct for that widget.  However, if there is a single reference to datetime in the JSON, then all dates that can be converted to datetime should be.  The way I do this manually is with a python function like this:  

```
def Convert_OHDSI_SQL_dates_to_datetime(str):
    import re

    step1 = re.sub(r'(?i)(?<!op_)(?<!observation_period_)(start|end|event|sort)_date\b',r'\1_datetime',str)
    step2 = re.sub(r'as cohort_definition_id, person_id, start_datetime, end_datetime',r'as cohort_definition_id, person_id, cast(start_datetime as date), cast(end_datetime as date)', step1)

    return step2
```

Can the circe-be function that does the SQL template generation be amended (or intercepted) to call this additional function (or its Java equivalent) before the final String is returned to the requestor.  From there, SqlRender should work fine on the syntax-specific translation.

chrisknoll:
Hi, @TomWhite-MedStar ,

I haven't dug deeply into this, as there's missing information in the PR description about what the changes are (in moderate detail), but currently has nothing yet.  Second issue is the entire notion of minute/second granularity in the cohort result, it is not clear what the impact on the external tools will be if there is a unit of time other than day.   

As far as using this for your own purposes, a branch or fork approach would work for  you to use this in your own environment if there is that need.  

As you might imagine that this cohort definition work is central to a lot of the OHDSI tool stack, we must be extremely careful what gets put into a formal release.   On the forums, I've observed support that it's a good idea (hour/minute granularity), but beyond that there are technical considerations (such as: sometimes the result query will put `_datetime` as an output column other times will be simply `_date`) that haven't been vetted, but we're making the decision on how that should work here.

So for starters, I think the PR author should include details on what the intended changes are that are being introduced in this PR (so that we have something to cross-check between the 'intended change' vs. the actual change in the 'lines of code changed').  This is important because we don't want to be left to reverse-engineer the lines of code that were changed to infer the intent of the author...it should be declared, and then we can judge the code changes based on that declared intent.

On small changes, it's pretty straight forward to declare what the intended changes of a PR is.  On something larger like this, we have new test cases, enums, date calculations (I think I saw in some cases that we intended to resolve the time diffs to seconds, but in certain places I see we're accounting for sometimes hours, sometimes minutes, sometimes seconds), so I just would like to know what I should be expecting when looking through the code changes before digesting all the lines of code changes.

Re: your approach to fix sql to date or date_time via string replace
I'm not a fan of this approach, I don't know if we can assume that date-time cols are always `_datetime`  I am thinking that it may be a cohort expression param that you specify the granularity of time for the output such that if you want it by hour, then you query all the entry events at the hour level, vs some other granularity.  This is definitely something that should be thought about at the design level, but we sorta jumped right into the implementation so there wasn't much time to talk about...but we have this implementation as a 'frame of reference' that we could raise technical issues, so that is also a fine perspective to reason from.

TomWhite-MedStar:
@chrisknoll , thanks for the lengthy reply.  I am presuming, but not sure, that this PR might be for work that Odysseus is doing for me at MedStar.  If so, I'm trying to get past a hurdle in that work.  They have working for months, and finding that the source  code is very complex - so I'm wanting to brainstorm whether there might be post-processing approach that is easier to implement and maintain.

chrisknoll:
Thanks for your understanding, @TomWhite-MedStar .   Although the codebase may be complicated, at the core it's a string-builder plus logic validator (through the checkers).   This means that, at the core, it's assembling the string through the various builder functions.   Introducing a notion of building the final string through a process of find/replaces might seem straightforward from an implementation perspective, but out of place from a functional perspective.   

So, lets get the description of what's in the PR in the description of this PR, resolve the conflicts, evaluate if the code is doing what it intended, and then we can make implementation recommendations based on findings.

PR #224, opened 2025-05-28
Remove effective_drug_dose field from Circe

YaroslavTir:
Following the discussion in [OHDSI/Atlas#2720](https://github.com/OHDSI/Atlas/issues/2720):

The effective_drug_dose field has been deprecated since OMOP CDM version 5.2.1 and is no longer part of the drug_exposure table. This PR removes the field from the cohort builder to align with the current CDM specification.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.ohdsi</groupId>
   <artifactId>circe</artifactId>
-  <version>1.12.0-SNAPSHOT</version>
+  <version>1.13.0-SNAPSHOT</version>
   <properties>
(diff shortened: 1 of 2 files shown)


chrisknoll:
I think we need to keep this for backwards compatibility?  As in: for those still on CDM 5.1, they may need to generate a cohort.

We started down a path of doing a SemVer range for a cohort definition and comparing it to the specific version of the target CDM to determine compatibility.  For example, using this attribute would mean that the semver range for the cohort definitin would be < 5.2,  We would use this semver range to check if a CDM on 5.1 would pass vs. a 5.2 would fail.

This feature will be more and more important as the CDM evolves.  I think we should talk on the WG call about how to tackle this problem.

chrisknoll (approved):

PR #234, opened 2026-08-30
Record cohort summary stats when a cohort has no inclusion rules

p-hoffmann:
Reported in OHDSI/Atlas3#299.

## Problem

A cohort with no inclusion rules generates normally and fills its cohort table, but records no base or final count anywhere.

`cohortInclusionAnalysis.sql` is the only thing that writes `cohort_summary_stats`, and in `generateCohort.sql` it sits behind:

```
{@generateStats != 0 & @ruleTotal != 0}?{
```

so the whole block is skipped when `@ruleTotal` is 0.

WebAPI's inclusion report reads that table and substitutes a default `Summary` when the query returns nothing:

```java
List<InclusionRuleReport.Summary> result = getSourceJdbcTemplate(source).query(...);
return result.isEmpty() ? new InclusionRuleReport.Summary() : result.get(0);
```

`Summary`'s counts are primitive `long`s, so that default is all zeros. The reporter generated a cohort, was told 1,331 records had been found, opened the Inclusion Report and was shown nothing at all.

## Why the guard is unnecessary

Nothing in the block requires a rule to exist:

| Step | With zero rules |
|---|---|
| `@inclusionRuleTable` | already emits `CREATE TABLE #inclusion_rules (rule_sequence int);`, so the table exists and is empty |
| `#best_events` | left joins an empty `#inclusion_events`, giving every event `rule_count` 0, and ranks on start date: the earliest event per person, which is what by-person mode means |
| `cohort_inclusion_result` | one row, mask 0, covering every event |
| `cohort_inclusion_stats` | selects `from #inclusion_rules`, which is empty, so no rows. Correct: no rules, nothing to report per rule |
| `cohort_summary_stats` | base is the qualifying event count; the "all rules matched" mask is `POWER(2,0)-1 = 0`, which selects exactly the row above, so final equals base |

Base and final both equal to the qualifying event count is the right answer for a cohort with no rule to drop anything, and it is what the report needs.

## Change

Drop `& @ruleTotal != 0` from the guard. The other `@ruleTotal` guard, on the final-cohort mask filter at line 29, is untouched and still correct.

## Cost

A rule-less generation now does the work the analysis block already did for every cohort that has rules, including building `#best_events` for the by-person mode. Flagging that explicitly in case it matters for very large rule-less cohorts.

## Tests

`testSummaryStatsWithoutInclusionRules` takes the existing `simpleInclusionRule` design, removes its single rule, generates against the embedded Postgres, and asserts:

- the cohort is non-empty, so the counts are worth measuring
- both modes (0 all events, 1 best event per person) record a summary row with a non-zero base count
- final equals base, since no rule can drop an event
- `cohort_inclusion_stats` stays empty, since there are no rules to report on

Before the change it fails on the first summary lookup with `EmptyResultDataAccessException: Incorrect result size: expected 1, actual 0`.

Full suite: **212 tests, 0 failures, 1 skipped**, BUILD SUCCESS.

## Related

Atlas3 side, OHDSI/Atlas3#310 stops the client presenting the absent summary as a report of zeros; this stops it being absent.

### DIFF
--- a/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
+++ b/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
@@ -80,7 +80,7 @@ delete from @results_database_schema.cohort_censor_stats where @cohort_id_field_
 @cohortCensoredStatsQuery
 -- END: Censored Stats
 }
-{@generateStats != 0 & @ruleTotal != 0}?{
+{@generateStats != 0}?{
 
 @inclusionRuleTable
 
--- a/src/test/java/org/ohdsi/circe/cohortdefinition/CohortGeneration_5_0_0_Test.java
+++ b/src/test/java/org/ohdsi/circe/cohortdefinition/CohortGeneration_5_0_0_Test.java
@@ -16,6 +16,7 @@ import org.dbunit.dataset.DataSetException;
 import org.dbunit.dataset.IDataSet;
 import org.dbunit.dataset.ITable;
 import org.dbunit.operation.DatabaseOperation;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.ohdsi.circe.AbstractDatabaseTest;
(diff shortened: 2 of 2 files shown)


chrisknoll:
Need to think about this a bit further, considering this code has worked this way for 10 years (and I'm fairly certain that we saw cohort counts this whole time):  the point of the gate is 'if you want generation stats and you have inclusion rules, then calculate stats).    All the tables you refrenced are dealing with entry-event / inclusion rule statistics.

On the front end, if your cohort definition does not have any inclusion rules we shouldn't show the inclusion rule report at all. i believe this is the 2.x behavior.

Do you want to confirm that behavior? I don't want to make changes in the circe-be package to support a UI concern.

-Chris

